PR #4, opened 2015-06-04
Okta 60284 hrishikesh add documentation

hrishikeshsathe-okta:
Please review this. I've added a few comments and would add more as I encounter more functions


lboyette-okta:
Looks good to me. Thanks!

PR #9, opened 2015-12-03
Added rate limit methods

lboyette-okta:
@tthompson-okta 

Added a RateLimitExceededException

Each httpResponse is stored until the next response
You can get the context using userClient.getRateLimitContext()

The supported methods are:
getNumRequestsRemaining()
getNextWindowUnixTime()
getNextWindowDateTime()
getRequestLimit()

### DIFF
--- a/src/main/java/com/okta/sdk/framework/ApiClient.java
+++ b/src/main/java/com/okta/sdk/framework/ApiClient.java
@@ -2,6 +2,7 @@ package com.okta.sdk.framework;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.okta.sdk.exceptions.ApiException;
+import com.okta.sdk.exceptions.RateLimitExceededException;
 import com.okta.sdk.exceptions.SdkException;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -25,6 +26,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 
 public abstract class ApiClient {
 
@@ -39,6 +41,7 @@ public abstract class ApiClient {
     protected String token;
     protected ApiClientConfiguration configuration;
     private final Logger LOGGER = LoggerFactory.getLogger(ApiClient.class);
+    private AtomicReference<HttpResponse> lastHttpResponseRef = new AtomicReference<HttpResponse>();
 
     ////////////////////////////////////////////
     // CONSTRUCTORS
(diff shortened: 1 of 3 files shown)


src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -39,6 +40,7 @@
     protected String token;
     protected ApiClientConfiguration configuration;
     private final Logger LOGGER = LoggerFactory.getLogger(ApiClient.class);
+    private HttpResponse lastHttpResponse;
```
lboyette-okta: My only question is whether this should be `static`, so the same RateLimitContext is available across all ApiClients

tthompson-okta: I don't think so, because different instances of ApiClient could technically be talking to different orgs (and thus have different rate limit statistics). I think this is good as a regular instance variable.

src/main/java/com/okta/sdk/framework/RateLimitContext.java
```diff
@@ -0,0 +1,58 @@
...
+        Header[] headers = httpResponse.getHeaders(headerName);
+        if (headers.length > 0) {
+            Header header = headers[0];
+            return header.getValue();
+        } else {
+            throw new SdkException("No " + headerName + " header");
+        }
+    }
+
+    private Long getHeaderValueLong(String headerName) throws Exception {
```
tthompson-okta: Would be better to have all these methods return a primitive `long` rather than an Object `Long`.

src/main/java/com/okta/sdk/framework/RateLimitContext.java
```diff
@@ -0,0 +1,58 @@
...
+        String headerString = getHeaderValueString(headerName);
+        try {
+            return Long.parseLong(headerString);
+        } catch (Exception e){
+            throw new SdkException("Error parsing " + headerName + " header");
+        }
+    }
+
+    public Long getNumRequestsRemaining() throws Exception {
```
tthompson-okta: Shouldn't we have detailed Javadoc on each of these explaining what they mean? It's not obvious just from reading the method names how exactly rate limiting works.

src/main/java/com/okta/sdk/framework/ApiClient.java
tthompson-okta: This should probably be an `AtomicReference<HttpResponse>` rather than just an HttpResponse. This way, multiple threads can make calls using this ApiClient and each one is guaranteed to see the latest value for `lastHttpResponse` (because to access it you will have to do `httpResponseRef.get()`).

src/main/java/com/okta/sdk/framework/RateLimitContext.java
lboyette-okta: Sounds good

src/main/java/com/okta/sdk/framework/ApiClient.java
lboyette-okta: Alright, good call.

src/main/java/com/okta/sdk/framework/RateLimitContext.java
lboyette-okta: We definitely should. I'll add them

tthompson-okta:
:ship:

PR #12, opened 2016-01-08
OKTA-77233

tthompson-okta:
Fix an issue where internal Okta dependencies leaked into the public POM file, breaking the build.

@jmaldonado-okta @senthilkumaran-okta @lboyette-okta @federations-okta

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,15 +9,8 @@
     <version>0.0.4-SNAPSHOT</version>
     <properties>
         <jackson.version>2.5.0</jackson.version>
-        <aperture-reporter.version>1.0.0-000079.000042.e302966</aperture-reporter.version>
-        <test.listener>do-not-listen</test.listener>
     </properties>
     <dependencies>
-        <dependency>
-            <groupId>com.okta.libraries.framework</groupId>
-            <artifactId>aperture-reporter</artifactId>
-            <version>${aperture-reporter.version}</version>
-        </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


jmaldonado-okta:
:shipit: I agree with this patch. We should revisit the original implementation for CI to avoid these dependencies and provide a robust solution.

mohitchugh-okta:
LGTM. Ship it!

pom.xml
```diff
@@ -14,11 +14,6 @@
     </properties>
     <dependencies>
         <dependency>
-            <groupId>com.okta.libraries.framework</groupId>
-            <artifactId>aperture-reporter</artifactId>
-            <version>${aperture-reporter.version}</version>
```
senthilkumaran-okta: The corresponding line, <aperture-reporter.version>1.0.0-000079.000042.e302966</aperture-reporter.version> could be removed.

You could also remove.
```
    <profiles>
        <profile>
            <id>ci-run</id>
            <properties>
                <test.listener>listener</test.listener>
            </properties>
        </profile>
    </profiles>
```
And then the build_and_test.sh script will have to see some changes, similar to how it is done here:
https://github.com/okta/org-creator/blob/master/scripts/build_and_test.sh

tthompson-okta:
@senthilkumaran-okta thanks, I've made the additional changes you suggested.

senthilkumaran-okta:
Thanks, the pom file changes looks good to go. #shipit :+1:

PR #14, opened 2016-01-20
Add ability to perform partial user updates

briguy202:
This pull is to expose the API's ability to submit partial user updates via POST calls, as instructed in the [Okta Documentation](http://developer.okta.com/docs/api/resources/users.html):

> Use the POST method to make a partial update and the PUT method to delete unspecified properties.

### DIFF
--- a/src/main/java/com/okta/sdk/clients/UserApiClient.java
+++ b/src/main/java/com/okta/sdk/clients/UserApiClient.java
@@ -163,6 +163,14 @@ public class UserApiClient extends JsonApiClient {
         return get(getEncodedPath("/%s/groups", userId), new TypeReference<List<UserGroup>>() { });
     }
 
+    public User partialUpdateUser(User params) throws IOException {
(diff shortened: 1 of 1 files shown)


briguy202:
Also, while considering this pull, could I also propose that the version in the POM have the 'SNAPSHOT' removed?

jpf:
Hi @briguy202 - thanks for this PR, I also note that PR #34 also requested similar changes. I've opened an internal ticket (OKTA-106322) for the team that manages this SDK.

Out of curiosity, I'm wondering:
- Are you using the methods that you added in production code?
- What is the use case that you needed to solve with this update?

briguy202:
Hi @jpf, in response to your questions:
* Yes, we are using the methods in production code.
* The use case we had was updating logins without needing to touch any other properties on the user's profile.

jmelberg-okta:
This enhancement looks to have been added in a [recent update (#47)](https://github.com/okta/oktasdk-java/pull/47/).

PR #15, opened 2016-02-08
OKTA-78337

larsjohansen-okta:
- Including Advanced User Search in the Users API client.
- Adding javadocs to the Users API client + general cleanup.
- Loosen up restrictions for the filter complexity.

Primary reviewers: @tthompson-okta @lboyette-okta @cwu-okta 
Secondary reviewers: @federations-okta @prestonchan-okta @sswaika-okta


src/main/java/com/okta/sdk/framework/FilterBuilder.java
```diff
@@ -204,24 +168,72 @@ public FilterBuilder lessThanOrEqual(DateTime value) {
...
+        filterBuilder.append(GREATER_THAN_OR_EQUAL_SIGN);
+        return this;
+    }
+
+    private FilterBuilder lessThan() {
+        filterBuilder.append(LESS_THAN_SIGN);
+        return this;
+    }
+
+    private boolean checkComplexity() {
```
tthompson-okta: I thought we decided to let the server enforce this. We never want to end up in a situation where the server loosens up its constraints (allows more operators) and then have the SDK explicitly precluding use of the new server abilities.

larsjohansen-okta: My understanding was to match the current restrictions, but I can remove it if you think it's better.

src/main/java/com/okta/sdk/clients/SessionApiClient.java
```diff
@@ -68,14 +64,22 @@ public Session createSessionWithSessionTokenAndAdditionalFields(String sessionTo
...
     public void clearSession(String sessionId) throws IOException {
         delete(getEncodedPath("/%s", sessionId));
     }
+
+    @Override
+    protected String getFullPath(String relativePath) {
+        return String.format("/api/v%d/sessions%s", this.apiVersion, relativePath);
+    }
+
 }
```
lboyette-okta: nit: newline

src/main/java/com/okta/sdk/clients/UserApiClient.java
```diff
@@ -334,37 +734,40 @@ public LoginCredentials changeRecoveryQuestion(String userId, String currentPass
...
-        return new PagedResults<User>(getUsersApiResponseAfterCursorWithLimit(after, limit));
+    @Override
+    protected String getFullPath(String relativePath) {
+        return String.format("/api/v%d/users%s", this.apiVersion, relativePath);
     }
 
-    public PagedResults<User> getUsersPagedResultsByUrl(String url) throws IOException {
-        return new PagedResults<User>(getUsersApiResponseByUrl(url));
-    }
 }
```
lboyette-okta: nit: I know this isn't part of your commit, but a newline here would be awesome

src/main/java/com/okta/sdk/framework/RateLimitContext.java
```diff
@@ -73,4 +51,28 @@ public DateTime getNextWindowDateTime() throws Exception {
...
+    private long getHeaderValueLong(String headerName) throws Exception {
+        String headerString = getHeaderValueString(headerName);
+        try {
+            return Long.parseLong(headerString);
+        } catch (Exception e){
+            throw new SdkException("Error parsing " + headerName + " header");
+        }
+    }
+
 }
```
lboyette-okta: nit: another for fun

src/main/java/com/okta/sdk/framework/FilterBuilder.java
tthompson-okta: I think it would be fine to doc the current restrictions on the developer site (if we don't already), but enforcing them in the client code is going too far IMO. I think we should remove it and let the server drive the business logic.

lboyette-okta:
nits + Trevor's comments then lgtm

src/main/java/com/okta/sdk/clients/UserApiClient.java
```diff
@@ -296,6 +593,109 @@ public LoginCredentials changeRecoveryQuestion(String userId, String currentPass
...
+    /**
+     * Get all users up to a limit using advanced search and cursor with paging info.
+     *
+     * @param  filterBuilder       The advanced search filter thats being used to match users.
+     * @param  limit               The max number of results returned.
+     * @param  after               The cursor that determines which users to return after.
+     * @return PagedResults<User>  The users with paging info.
+     * @throws IOException         If an input or output exception occurred.
+     */
+    public PagedResults<User> getUsersPagedResultsWithAdvancedSearchAndLimitAndAfterCursor(FilterBuilder filterBuilder, int limit, String after) throws IOException {
```
tthompson-okta: I'd just like to reiterate a couple design principles (from the API manifesto) to keep in mind with any SDK change:

* API Should Be As Small As Possible But No Smaller
* When in doubt leave it out; you can always add but you can never remove
* Public APIs are forever - one chance to get it right
* Minimize Accessibility of Everything
* Names Matter–the API is a Little Language

Here, the name `getUsersPagedResultsWithAdvancedSearchAndLimitAndAfterCursor()` is very descriptive, but it's almost unwieldy. Also, do we want to call the new functionality "Advanced Search" here or just "Search"? From the client's perspective, there isn't anything "advanced" about it; its just SCIM filter no different than the one it's already using via the `filter` parameter.

src/main/java/com/okta/sdk/clients/SessionApiClient.java
larsjohansen-okta: nit: nit: That's actually intentional, see for example the implementation of the String class.

src/main/java/com/okta/sdk/clients/UserApiClient.java
larsjohansen-okta: I'm a bit confused, why would we want a new line after the closing bracket?

src/main/java/com/okta/sdk/framework/RateLimitContext.java
larsjohansen-okta: nit: nit: See my answer above.

src/main/java/com/okta/sdk/framework/FilterBuilder.java
larsjohansen-okta: Ok, sounds good to me, in that way we will never get into a situation where we have more strict constraints on the client than the server, which would be really bad.

src/main/java/com/okta/sdk/clients/UserApiClient.java
larsjohansen-okta: We've had the convention of calling the q parameter the search functionality (both in code and on developer.okta.com, see http://developer.okta.com/docs/api/resources/users.html#list-users-with-search) so I thought advanced search was the official term we used for this. I've also seen that in the SDK we're using the term query (```getUsersWithQuery```) more than search, so there's already a base for confusion when reading the SDK and the documentation on developer.okta.com. I thought that it would be a good idea to be consistent with the docs now that we have the chance, instead of making it confusing saying that the new functionality is called search but it's not the search that we document online, it's what we call advanced search online.

Regarding the other part of the name for this method, I agree it is very verbose; I was first thinking of using the name ```getUsersPagedResultsWithAdvancedSearch``` and overload this method with all 3 parameter combinations (search, search + limit, search + limit + cursor). I didn't end up doing that to follow the convention we use with filtering, which is spelling out the parameters in the name of the function. So I followed consistency in this case.

Let me know what you think we should call it; I think we should get rid of parameter names in the method name, and instead assume they read the javadocs and understands. I would then call them all ```getUsersPagedResultsWithAdvancedSearch(...)```.

src/main/java/com/okta/sdk/clients/UserApiClient.java
larsjohansen-okta: Fixed.

src/main/java/com/okta/sdk/clients/UserApiClient.java
tthompson-okta: Looking at this again and thinking through it more, I like it the way you have it. Thanks for the explanation as well.

tthompson-okta:
:rocket:

README.md
```diff
@@ -75,44 +65,52 @@ user.setProfile(userProfile);
...
-List<User> users = userClient.getUsersWithFilter(filterBuilder);
+List<User> users = userApiClient.getUsersWithFilter(filterBuilder);
+
+// 3. Advanced search provides the option to filter on any user profile attribute, any custom defined
+// profile attribute, as well as the following top-level attributes: id, status, created, activated, 
+// statusChanged and lastUpdated. The advanced search performs a case insensitive filter against all fields
+// specified in the search parameter. Note that the results might not yet be up to date, as the most up to date
+// data can be delayed up to a few seconds, so use for convenience.
+FilterBuilder filterBuilder = new FilterBuilder("profile.flightNumber eq \"A415\"");
+List<User> users = userApiClient.getUsersWithSearch(filterBuilder);
```
cwu-okta: Do we need to call this a beta feature?

PR #18, opened 2016-03-24
Close the connection when a parsing exception is thrown

michaelpage:
If a parsing exception is thrown while trying to read the HTTP response, the `HttpClient` connection is not closed properly. Once all the connections in the pool have reached this state, the client is unable to send any requests to the Okta API until the service is restarted. This issue was found during a penetration test and rendered our service unresponsive. I re-worked the `JsonApiClient.unmarshall()` method to attempt to close the connection in a finally block to prevent this issue from occurring.


bvbsdvy:
The problem described here affected us too and caused our app to go down in production. We had to restart our server to recover.

joelfranusic-okta:
Hi @michaelpage thank you very much for this PR and @bvbsdvy, thank you for confirming that you had this issue too. I've opened a ticket (OKTA-106307) for this issue with the team that maintains this code.

bdemers:
@michaelpage sorry this took so long!

PR #23, opened 2016-04-22
Intermittent API call failures in OKTA SDK Resolves: OKTA-86647 Bacon: test

ashishgupta-okta:
Fixed intermittent API call failures due to Apache HttpClient. Added a connection manager and retry strategy per Apache documentation.

Resolves : OKTA-86647
Primary Reviewer : @royalchan-okta @tthompson-okta 
Reviewers: @federations-okta

### DIFF
--- a/src/main/java/com/okta/sdk/framework/ApiClient.java
+++ b/src/main/java/com/okta/sdk/framework/ApiClient.java
@@ -15,6 +15,7 @@ import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpPut;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.client.StandardHttpRequestRetryHandler;
 import org.apache.http.message.BasicHeader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -29,6 +30,8 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
 public abstract class ApiClient {
+
+    public static final int RETRY_COUNT = 3;
 
     public static final String AFTER_CURSOR = "after";
     public static final String LIMIT = "limit";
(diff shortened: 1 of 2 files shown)


cwu-okta:
Nit: can we clean up the commit message? It is going to be seen publicly so we should have well formed messages. Since this is for the Java SDK repo we do not need to prefix the message to indicate this is for the Java SDK. A suggestion which came to mind:

```FIx intermittent API call failures due to Apache HttpClient. Added a connection manager and retry strategy per Apache documentation.

Resolves: OKTA-86647```

File: src/main/java/com/okta/sdk/framework/ApiClient.java [hunk 1]
```diff
@@ -45,18 +48,23 @@
...
-                        .setUserAgent("OktaSDKJava_v" + Utils.getSdkVersion())
-                        .disableCookieManagement().build(),
-                config);
-    }
-
-    protected ApiClient(HttpClient httpClient, ApiClientConfiguration config) {
-        this.httpClient = httpClient;
+        PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
+        connectionManager.closeExpiredConnections();
+        connectionManager.closeIdleConnections(5, TimeUnit.SECONDS); // Our TTL is 5 seconds
```
cwu-okta: Nit: can we make this a constant?

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -45,18 +48,23 @@
...
+
+        // The connections automatically expire/stale in 5000ms by default so we do not need to set the connection time out.
+        // When the connection expires/goes stale and sometimes when the httpClient is not aware, the next subsequent call fails which make the client
+        // realize that the connection has expired. The request can then be retried by the httpClient.
+        //https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html Section 2.5
```
cwu-okta: Please do not link to the documentation. We can talk more about this if you have questions.

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -45,18 +51,19 @@
...
 
     public ApiClient(ApiClientConfiguration config) {
-        this(HttpClientBuilder.create()
-                        .setUserAgent("OktaSDKJava_v" + Utils.getSdkVersion())
-                        .disableCookieManagement().build(),
-                config);
-    }
+        PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager();
+        connectionManager.closeExpiredConnections();
+        connectionManager.closeIdleConnections(TTL, TimeUnit.SECONDS);
```
bvbsdvy: I don't think calling `connectionManager.closeExpiredConnections()` or `connectionManager.closeIdleConnections(TTL, TimeUnit.SECONDS)` will do anything useful at this point. According to the [PoolingHttpClientConnectionManager Javadoc](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.html#closeIdleConnections(long,%20java.util.concurrent.TimeUnit)), these methods close idle or expired connections at the time they are called. But right after creating the `PoolingHttpClientConnectionManager` there are not even any connections yet.

Normally these methods would be called periodically from a background thread (see section 8 of this [HttpClient guide](http://www.baeldung.com/httpclient-connection-management).

Also I don't think it's necessary to call both: the documentation for `closeIdleConnections` states "All expired connections will also be closed."

Anyway the easiest way to set a persistent connection TTL appears to be [`HttpClientBuilder.setConnectionTimeToLive`](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html#setConnectionTimeToLive(long,%20java.util.concurrent.TimeUnit)).

ashishgupta-okta: When you say periodically, what should be that period? What if the server closes the connection in that period but the client still uses that connection? The request will then fail. Retrying the requests seems like the safest approach to serve requests.  
Also, I tried the setConnectionTimeToLive and it did not solve the issue.

ashishgupta-okta: Also, I dont think we have to use the connectionManager.closeExpiredConnections() or connectionManager.closeIdleConnections(TTL, TimeUnit.SECONDS), retrying requests serves the purpose

royalchan-okta:
🚢

PR #28, opened 2016-05-27
Added ability for createUser() to also specify groupIds that the new …

brianzuzga-okta:
…user should immediately be added to.  This was done by adding the groupIds array to the User object.  This array is only read by the create user API call at present, and it is not planned that it will ever be returned in any responses.

### DIFF
--- a/src/main/java/com/okta/sdk/models/users/User.java
+++ b/src/main/java/com/okta/sdk/models/users/User.java
@@ -6,6 +6,7 @@ import com.okta.sdk.framework.ApiObject;
 import com.okta.sdk.models.links.LinksUnion;
 import org.joda.time.DateTime;
 
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


brianzuzga-okta:
@federations-okta I was told this was the right alias for review.

brianzuzga-okta:
The API change that this SDK change supports was merged last week.  It would be good to get this in soon.

cwu-okta:
@brianzuzga-okta Can you indicate what the exact commit message will be for this merge? Please be sure to also include the Jira number. We need crisp, clear commit message for public repos.

brianzuzga-okta:
@joelfranusic-okta Here is the updated commit message with the (newly created) JIRA.  I misinterpreted the instructions on the wiki that said not to include internal references in public SDK commits to mean JIRAs as well.

PR #34, opened 2016-09-07
Expose endpoints for partial user updates

Soniaroxana:


jpf:
Hi @Soniaroxana, thanks for this pull request! I've created a ticket (OKTA-106322) for the team that manages this code to look into adding support for partial user updates. 

Out of curiosity, I'm wondering:
- Are you using the methods that you added in production code?
- What is the use case that you needed to solve with this update?

Soniaroxana:
@jpf the methods that I created are simply a wrapper of the REST API endpoints that you(Okta) released a couple of months ago, but did not implement in the Java SDK.

The use case is this:
- every time you want to make an user update, with the current SDK method for update you need to pass in a full user object (with ALL the properties), otherwise the properties that are not present will be overwritten with null or in some cases, the update will fail. This is because of how the Okta REST API endpoint for user update using the PUT method is implemented. This is inconvenient and dumb if you just want to update a single user property like First or Last name.
- a couple of months ago Okta created a new endpoint for user update using the POST method, which allows you to do partial user updates, like updating just one field, aka First/Last name, but it was not implemented in the Java SDK. Therefore, I created this pull request 

yes, we are using these endpoints in production code, we created our own branch of the Okta Java SDK.

jmelberg-okta:
Hi @Soniaroxana! Thanks for your pull request.

Looks like **partial user updates** were addressed in a [recent push (#47)](https://github.com/okta/oktasdk-java/pull/47/).

However, before we can move forward with reviewing/merging your PR, please take a moment to read and sign a Contributor License Agreement (CLA).

Please visit [developer.okta.com/cla](developer.okta.com/cla) to sign.

Then, reply here (“`Signed`”) so we can proceed. Thanks!

bdemers:
@Soniaroxana closed until signed.
Please just comment and we will open this back up

PR #38, opened 2016-12-15
Automatic proxy detection

AlainODea:
Use Java's ProxySelector to detect proxy JRE proxy settings.


AlainODea:
I need to make this work for the case with no proxy config. Bear with me.

AlainODea:
Recreated as #39.

PR #39, opened 2016-12-15
:bug: FIX #25: Automatic proxy detection

### LINKED ISSUE #25: "No route to host" when performing operation createUser behind proxy
Hello,

When i try to connect to OKTA, i get this error : 

> [main] INFO com.okta.sdk.framework.ApiClient - POST https://subdomain.okta.com/api/v1/users?activate=true 
> May 3, 2016 4:40:04 PM org.apache.http.impl.execchain.RetryExec execute
> INFO: I/O exception (java.net.NoRouteToHostException) caught when processing request to {s}->https://subdomain.okta.com:443: No route to host
> May 3, 2016 4:40:04 PM org.apache.http.impl.execchain.RetryExec execute
> INFO: Retrying request to {s}->https://subdomain.okta.com:443

indeed, i try to communicate whith OKTA behind a proxy server

i used JVM properties to set my proxy adresses

```
            System.setProperty("http.proxyHost", "ipadress");
            System.setProperty("http.proxyPort", "8080");
            System.setProperty("https.proxyHost", "ipadress");
            System.setProperty("https.proxyPort", "8080");
```

Apparently these lines of code are not working !

is there any way to set proxy properties using the java API ?

NB : Downloading the file from the same subdomain https://subdomain.okta.com using Java is working with this code :

```
    public static void download(String address, String localFileName) {
        OutputStream out = null;
        URLConnection conn = null;
        InputStream in = null;

    System.setProperty("http.proxyHost", "ipadress");
    System.setProperty("http.proxyPort", "8080");
    System.setProperty("https.proxyHost", "ipadress");
    System.setProperty("https.proxyPort", "8080");

        try {
            URL url = new URL(address);
            out = new BufferedOutputStream(new FileOutputStream(localFileName));
            conn = url.openConnection();
            in = conn.getInputStream();
            byte[] buffer = new byte[1024];

            int numRead;
            long numWritten = 0;

            while ((numRead = in.read(buffer)) != -1) {
                out.write(buffer, 0, numRead);
                numWritten += numRead;
            }

            System.out.println(localFileName + "\t" + numWritten);
        } 
        catch (Exception exception) { 
            exception.printStackTrace();
        } 
        finally {
            try {
                if (in != null) {
                    in.close();
                }
                if (out != null) {
                    out.close();
                }
            } 
            catch (IOException ioe) {
            }
        }
    }
```

You will find below the code returning "No route to host" error:

```
package okta;

import java.io.IOException;

import com.okta.sdk.clients.UserApiClient;
import com.okta.sdk.framework.ApiClientConfiguration;
import com.okta.sdk.models.users.User;

public class Main {

        public static void main(String[] args) {
        System.out.println( "Hello World!" );

        System.out.println( "Setting proxy..." );
        System.setProperty("http.proxyHost", "ipadress");
        System.setProperty("http.proxyPort", "8080");
        System.setProperty("https.proxyHost", "ipadress");
        System.setProperty("https.proxyPort", "8080");

        System.out.println( "Setting config Okta..." );
        ApiClientConfiguration oktaSettings = new ApiClientConfiguration(
                "https://subdomain.okta.com",
                "test");

        UserApiClient userApiClient = new UserApiClient(oktaSettings);

     try {
            System.out.println( "Calling Okta..." );
            User newUser = userApiClient.createUser(
                             "First",
                             "Last",
                             "login@example.com",
                             "email@example.com",
                             true);
            System.out.println( "End calling" );
        } catch (IOException e) {
            System.out.println( "KO" );
            e.printStackTrace();
        }

        System.out.println( "End" );
    }

}
```

Thank you in advance.

AlainODea:
Use Java's ProxySelector to detect proxy JRE proxy settings.

### DIFF
--- a/src/main/java/com/okta/sdk/framework/ApiClient.java
+++ b/src/main/java/com/okta/sdk/framework/ApiClient.java
@@ -18,6 +18,7 @@ import com.okta.sdk.exceptions.RateLimitExceededException;
 import com.okta.sdk.exceptions.SdkException;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
+import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
 import org.apache.http.StatusLine;
 import org.apache.http.client.HttpClient;
@@ -26,14 +27,19 @@ import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpPut;
 import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.conn.routing.HttpRoutePlanner;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.StandardHttpRequestRetryHandler;
+import org.apache.http.impl.conn.DefaultProxyRoutePlanner;
+import org.apache.http.impl.conn.DefaultRoutePlanner;
 import org.apache.http.message.BasicHeader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.net.Proxy;
+import java.net.ProxySelector;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
@@ -64,9 +70,20 @@ public abstract class ApiClient {
     ////////////////////////////////////////////
 
     public ApiClient(ApiClientConfiguration config) {
+        Proxy proxy = ProxySelector.getDefault().select(URI.create(config.getBaseUrl())).iterator().next();
+        HttpRoutePlanner routePlanner;
+        if (Proxy.Type.HTTP.equals(proxy.type())) {
+            URI proxyUri = URI.create(proxy.type() + "://" + proxy.address());
+            HttpHost proxyHost = new HttpHost(proxyUri.getHost(), proxyUri.getPort(), proxyUri.getScheme());
+            routePlanner = new DefaultProxyRoutePlanner(proxyHost);
+        } else {
+            routePlanner = new DefaultRoutePlanner(null);
+        }
         StandardHttpRequestRetryHandler requestRetryHandler = new StandardHttpRequestRetryHandler(RETRY_COUNT, true);
         HttpClient client = HttpClientBuilder.create().setRetryHandler(requestRetryHandler)
-                .setUserAgent("OktaSDKJava_v" + Utils.getSdkVersion()).disableCookieManagement().build();
+                .setUserAgent("OktaSDKJava_v" + Utils.getSdkVersion()).disableCookieManagement()
+                .setRoutePlanner(routePlanner)
+                .build();
 
         this.httpClient = client;
         this.baseUrl = config.getBaseUrl();


AlainODea:
Fixes #25.

joelfranusic-okta:
Hi Alain, thanks for your PR! Would you be able to sign the [Okta CLA](http://developer.okta.com/cla) for us so that we could merge your PR?

AlainODea:
@joelfranusic-okta thank you. I have emailed my signed Okta Individual Contributor License Agreement and referenced this pull request in the message :)

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -16,28 +16,25 @@
 import com.okta.sdk.exceptions.ApiException;
 import com.okta.sdk.exceptions.RateLimitExceededException;
 import com.okta.sdk.exceptions.SdkException;
-import org.apache.http.Header;
-import org.apache.http.HttpEntity;
-import org.apache.http.HttpResponse;
-import org.apache.http.StatusLine;
+import org.apache.http.*;
```
yuliu-okta: Please do not use wildcard import

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -16,28 +16,25 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.net.URLEncoder;
+import java.net.*;
```
yuliu-okta: same as above

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -64,9 +61,20 @@
     ////////////////////////////////////////////
 
     public ApiClient(ApiClientConfiguration config) {
+        Proxy proxy = ProxySelector.getDefault().select(URI.create(config.getBaseUrl())).iterator().next();
+        URI proxyUri = URI.create(proxy.type() + "://" + proxy.address());
```
yuliu-okta: What if there is no proxy is configured, proxy is null?

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -64,9 +61,20 @@
...
+        HttpRoutePlanner routePlanner;
+        if (Proxy.Type.HTTP.equals(proxy.type())) {
```
yuliu-okta: The proxy selector should already select the proxy based on the URI, why here check the protocol again?

src/main/java/com/okta/sdk/framework/ApiClient.java
AlainODea: Sorry. Bad defaults in IntelliJ. I'll fix that.

src/main/java/com/okta/sdk/framework/ApiClient.java
AlainODea: This is intentional. Commons HttpClient only supports HTTP proxies so if there is a non-HTTP proxy (say SOCKS) it will fallback on connecting directly.

src/main/java/com/okta/sdk/framework/ApiClient.java
AlainODea: Okay. That's a good question. Here's the Javadoc for [**ProxySelector#select(URI)**](https://docs.oracle.com/javase/8/docs/api/java/net/ProxySelector.html#select-java.net.URI-):

> **select**
>
> public abstract List<Proxy> select(URI uri)
> Selects all the applicable proxies based on the protocol to access the resource with and a destination address to access the resource at. The format of the URI is defined as follow:
>
>  * http URI for http connections
>  * https URI for https connections
>  * socket://host:port for tcp client sockets connections
>
> **Parameters:**
> uri - The URI that a connection is required to
> 
> **Returns:**
> a List of Proxies. Each element in the the List is of type Proxy; when no proxy is available, the list will contain one element of type Proxy that represents a direct connection.
>
> **Throws:**
> IllegalArgumentException - if the argument is null

I think we're covered there. In the case where there is no proxy it will return a single element list with a DIRECT proxy.

yuliu-okta: Ok, can we move this line into the if block at line 76, seems we do not need to create an URI for the DIRECT/SOCKS cases.

AlainODea: Good idea. I'll do that :)

AlainODea:
Now squashed and rebased on master.

lboyette-okta:
This should make it into maven later this week or early next week.

PR #42, opened 2017-01-09
:seedling: Allows passing factorType to forgotPassword

lboyette-okta:
Resolves: OKTA-106319

@rchild-okta @ravikanodia-okta

### DIFF
--- a/src/main/java/com/okta/sdk/clients/AuthApiClient.java
+++ b/src/main/java/com/okta/sdk/clients/AuthApiClient.java
@@ -39,6 +39,11 @@ public class AuthApiClient extends JsonApiClient {
     public static final String OLD_PASSWORD = "oldPassword";
     public static final String NEW_PASSWORD = "newPassword";
 
+    public enum FactorType {
+        EMAIL,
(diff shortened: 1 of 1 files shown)


src/main/java/com/okta/sdk/clients/AuthApiClient.java
```diff
@@ -141,9 +146,16 @@ public AuthResult resetPassword(String stateToken, String relayState, String new
     }
 
     public AuthResult forgotPassword(String username, String relayState) throws IOException {
+        return forgotPassword(username, null, relayState);
+    }
+
+    public AuthResult forgotPassword(String username, FactorType factorType, String relayState) throws IOException {
```
ravikanodia-okta: Can you put factorType last, minimizing the difference in prototypes?

lboyette-okta: I think we'll ultimately make the relayState optional, since it's not used by everyone. So considering the addition of future methods, I think it makes sense to keep it here.

PR #44, opened 2017-01-12
:bug: Log tests depend on UTC system timezone

AlainODea:


AlainODea:
If I run the tests on my Macbook with a local timezone they fail. If I run them with a UTC timezone set they pass.

To address this I modified the tests to construct DateTime objects with UTC timezone explicitly set. I'm not sure if this masks a more serious underlying issue.

drj101687:
I just tried your change and it resolved the problem for me as well. How did you get around the javadoc plugin error?

`Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.9.1:jar (attach-javadocs) on project okta-sdk`
Then the build prints out a lot of errors about missing JavaDoc keywords. example:
`[ERROR] /Users/darrin.johnson/Dev/oktasdk-java/src/main/java/com/okta/sdk/models/apps/AppInstance.java:270: warning: no @param for val

[ERROR] public void setSettings(Settings val) {`

AlainODea:
@drj101687 the javadoc goal still fails, but the SNAPSHOT JAR gets built which was my interim goal with this :)

I get an error about not setting **JAVA_HOME** in my case. That was a blast from the past. I didn't realize that was still a thing.

AlainODea:
@lboyette-okta @guyzehavi-okta I think this could make the out of the box build experience better for contributors. Before this fix the tests involved fail unless the local system's timezone is set to UTC. My change allows contributors to have these tests pass even if their timezone is not set to UTC.

PR #53, opened 2017-02-01
:seedling: Add support for custom headers

jmelberg-okta:


src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -233,6 +233,13 @@ protected void setHeaders(HttpUriRequest httpUriRequest) throws IOException {
         // Content type and accept header are determined by the child class
         setContentTypeHeader(httpUriRequest);
         setAcceptHeader(httpUriRequest);
+
+        // Walk through custom headers and set
+        Map<String, String> headers = this.configuration.getHeaders();
+        for (Map.Entry<String, String> header : headers.entrySet()) {
+            Header authHeader = new BasicHeader(header.getKey(), header.getValue());
```
lboyette-okta: We're not guaranteed that this is an authHeader, so maybe a better name is `extraHeader` or something similar

src/main/java/com/okta/sdk/framework/ApiClient.java
```diff
@@ -233,6 +233,13 @@ protected void setHeaders(HttpUriRequest httpUriRequest) throws IOException {
         // Content type and accept header are determined by the child class
         setContentTypeHeader(httpUriRequest);
         setAcceptHeader(httpUriRequest);
+
+        // Walk through custom headers and set
+        Map<String, String> headers = this.configuration.getHeaders();
+        for (Map.Entry<String, String> header : headers.entrySet()) {
```
lboyette-okta: It's possible that headers is null, so we should check for that

PR #56, opened 2017-02-25
Add UserApiClient tests

lboyette-okta:
@rchild-okta 
@jmelberg-okta


package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@okta/oktasdk-java",
+  "version": "0.0.1",
```
lboyette-okta: Not sure if we care about this version number matching the pom.xml. I don't think it matters, and it just adds to what we have to think about when we bump the version of the artifact.

Also curious if we should do `private: true` to prevent accidental pushes to npm

lboyette-okta: Not sure if we care about this version number matching the pom.xml. I don't think it matters. It just adds to what we have to think about when we bump the version of the artifact.

Also curious if we should do `private: true` to prevent accidental pushes to npm

src/main/java/com/okta/sdk/clients/UserApiClient.java
```diff
@@ -78,6 +78,9 @@ public UserApiClient(ApiClientConfiguration config) {
      * @throws IOException  If an input or output exception occurred.
      */
     public List<User> getUsersWithLimit(int limit) throws IOException {
+        if (limit <= 0) {
```
jmelberg-okta: The API will return users when `limit` is `-1`. Anything `< -1` throws a `400`. Should we just check the zero case and let Okta return the error message?

PR #59, opened 2017-03-06
Updates readme to remove while(true)

dogeared:
:memo: Updates readme to remove `while(true)`
:art: Updates readme to remove `while(true)`


joelfranusic-okta:
Thanks!

PR #60, opened 2017-03-06
Adds UserGroup API tests

lboyette-okta:
This adds tests for our UserGroupApiClient using our mock-okta client.

@rchild-okta @magizh-okta @jmelberg-okta


src/test/java/TestConfig.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    private MockOkta mockOkta;
+
+    public MockOkta getMockOkta() {
+        return this.mockOkta;
+    }
+
+    public void setMockOkta(MockOkta mockOkta) {
```
jmelberg-okta: This POJO class is used [here](https://github.com/okta/oktasdk-java/blob/c2be4ab1056c83f17fb6fd48c89b144789ad51ca/src/test/java/UserApiClientTest.java#L28).

src/main/java/com/okta/sdk/clients/UserGroupApiClient.java
```diff
@@ -49,6 +49,9 @@ public UserGroupApiClient(ApiClientConfiguration config) {
     }
 
     public List<UserGroup> getUserGroupsWithLimit(int limit) throws IOException {
+        if (limit == -1) {
```
lboyette-okta: There is a default limit, which we use in the `UserApiClient`, but not here. I'll add a method that doesn't take a limit.

src/test/java/TestConfig.java
lboyette-okta: The `setMockOkta` method is required for Jackson to deserialize JSON into the class.

PR #62, opened 2017-03-22
:memo: Adds documentation comments to Models

jmelberg-okta:
:memo: Adds documentation comments to `Models` for JavaDoc generation


jmelberg-okta:
Notable changes:
- Verified all setter methods pass in an argument name `val` for consistency.
- All first-line comment strings ends with period.
- Provide links to object types in getter/setter methods
- Add newlines to the end of files

bdemers:
Looks like the current tests are expecting something on port 3000, we will need to fix that, (or make the test start up the service directly), but that is a different topic.

You should be able to see all of the lint errors by running this command:
`mvn -P\!java8-doclint-disabled javadoc:javadoc`

Basically, it will disable the profile `java8-doclint-disabled` using the bang and just parse/generate the javadoc

Good start!!

bdemers:
Sweet, If by some crazy way, you get all of these javadoc errors fixed, be sure to remove `java8-doclint-disabled` profile, so it stats that way.

PR #64, opened 2017-04-10
Authentication using activation token

nikolaystoilov-okta:
Added authentication method using activation token. 
Extended authentication response with authentication token.

Reviewers:
@srinivasanagandla-okta @edjohns-okta @yuliu-okta @rchild-okta 

Additional reviewers:
@authteam-okta 

Resolves: OKTA-118829

Bacon: test

### DIFF
--- a/src/main/java/com/okta/sdk/clients/AuthApiClient.java
+++ b/src/main/java/com/okta/sdk/clients/AuthApiClient.java
@@ -38,6 +38,7 @@ public class AuthApiClient extends JsonApiClient {
     public static final String RECOVERY_TOKEN = "recoveryToken";
     public static final String OLD_PASSWORD = "oldPassword";
     public static final String NEW_PASSWORD = "newPassword";
+    public static final String TOKEN = "token";
 
     public enum FactorType {
         EMAIL,
@@ -111,6 +112,19 @@ public class AuthApiClient extends JsonApiClient {
         params.put(RELAY_STATE, relayState);
         params.put(CONTEXT, context);
         return post(getEncodedPath("?response_type=%s&force_mfa=%s", responseType, String.valueOf(forceMFA)), params, new TypeReference<AuthResult>() { });
+    }
+
+    /**
+     * Authenticate into organization.
+     *
+     * @param  token {@link String}                 activationToken to use for authentication.
+     * @return {@link AuthResult}                   Result of the authentication transaction.
+     * @throws IOException                          If an input or output exception occurred.
+     */
+    public AuthResult authenticate(String token) throws IOException {
+        Map<String, Object> params = new HashMap<String, Object>();
+        params.put(TOKEN, token);
+        return post(getEncodedPath(""), params, new TypeReference<AuthResult>() { });
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/okta/sdk/clients/AuthApiClient.java
```diff
@@ -114,6 +115,19 @@ public AuthResult authenticate(String username, String password, String relaySta
...
+     * Authenticate into organization.
+     *
+     * @param  token {@link String}                 activationToken to use for authentication.
+     * @return {@link AuthResult}                   Result of the authentication transaction.
+     * @throws IOException                          If an input or output exception occurred.
+     */
+    public AuthResult authenticate(String token) throws IOException {
+        Map<String, Object> params = new HashMap<String, Object>();
+        params.put(TOKEN, token);
+        return post(getEncodedPath(""), params, new TypeReference<AuthResult>() { });
```
srinivasanagandla-okta: nit: Feels little odd to have the method take empty string as the value. Is there way to make it cleaner?

src/main/java/com/okta/sdk/clients/AuthApiClient.java
```diff
@@ -114,6 +115,19 @@ public AuthResult authenticate(String username, String password, String relaySta
...
 
     /**
+     * Authenticate into organization.
+     *
+     * @param  token {@link String}                 activationToken to use for authentication.
+     * @return {@link AuthResult}                   Result of the authentication transaction.
+     * @throws IOException                          If an input or output exception occurred.
+     */
+    public AuthResult authenticate(String token) throws IOException {
+        Map<String, Object> params = new HashMap<String, Object>();
```
mraible: Should be able to change this to `Map<String, Object> params = new HashMap<>();`.

http://www.javaworld.com/article/2074080/core-java/jdk-7--the-diamond-operator.html

src/main/java/com/okta/sdk/clients/AuthApiClient.java
```diff
@@ -114,6 +115,19 @@ public AuthResult authenticate(String username, String password, String relaySta
     }
 
     /**
+     * Authenticate into organization.
+     *
+     * @param  token {@link String}                 activationToken to use for authentication.
+     * @return {@link AuthResult}                   Result of the authentication transaction.
+     * @throws IOException                          If an input or output exception occurred.
+     */
+    public AuthResult authenticate(String token) throws IOException {
```
rchild-okta: Should we call this authenticateWithActivationToken, like we have authenticateWithFactor below? Also as a nit, can you change the description from "Authenticate into organization" to "Authenticate with activation token"?

src/main/java/com/okta/sdk/clients/AuthApiClient.java
nikolaystoilov-okta: This will require to change the sdk version of the project. (now is 1.5) This should be in different change.

srinivasanagandla-okta:
LGTM. 🚢 

Get a ship from Reman as well before you merge.

src/main/java/com/okta/sdk/clients/AuthApiClient.java
bdemers: +1 for changing the JDK version along with this change. Otherwise it would still take two additional commits to fix this. (or tweak the pom for 1.7 source/target, and rebase this on top)

bdemers:
Dupe of #72

PR #65, opened 2017-04-11
Fix the build script to start/stop the mock server.

nikolaystoilov-okta:
Added code to install npm dependancies.
Added code to start the mock server before the test and stop it after the test.
Changed the script target from deploy to package.

Reviewer:
    @davidjoyce-okta @lboyette-okta 

Bacon: test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,7 @@
         <testng.version>6.8.17</testng.version>
         <org.hamcrest.version>1.3</org.hamcrest.version>
         <com.google.guava.version>18.0</com.google.guava.version>
+        <frontend-maven-plugin.version>1.3</frontend-maven-plugin.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
     <dependencies>
@@ -111,6 +112,53 @@
     </profiles>
     <build>
         <plugins>
+            <plugin>
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
+                <version>${frontend-maven-plugin.version}</version>
+                <configuration>
+                    <nodeVersion>v5.6.0</nodeVersion>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>install node and npm</id>
+                        <goals>
+                            <goal>install-node-and-npm</goal>
+                        </goals>
+                        <phase>initialize</phase>
+                    </execution>
+                    <execution>
+                        <id>npm configure</id>
+                        <goals>
+                            <goal>npm</goal>
+                        </goals>
+                        <phase>initialize</phase>
+                        <configuration>
+                            <arguments>config set strict-ssl false</arguments>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>npm install</id>
+                        <goals>
+                            <goal>npm</goal>
+                        </goals>
+                        <phase>initialize</phase>
+                        <configuration>
+                            <arguments>install</arguments>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>start mock server</id>
+                        <goals>
+                            <goal>npm</goal>
+                        </goals>
+                        <phase>generate-resources</phase>
+                        <configuration>
+                            <arguments>run pretest</arguments>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
--- a/scripts/build_and_test.sh
+++ b/scripts/build_and_test.sh
@@ -15,9 +15,9 @@ BUILD_AND_TEST_TEST_SUITE_ID=771f3f55-7db9-11e5-80ff-22000b5181e9
 start_test_suite ${BUILD_AND_TEST_TEST_SUITE_ID}
  
 function build_and_test() {
-  ${MAVEN} clean deploy -s "${MAVEN_SETTINGS}"
+    ${MAVEN} clean package -s "${MAVEN_SETTINGS}"
 }
- 
+
 if ! build_and_test; then
   fail_test_suite
   exit 1


davidjoyce-okta:
It fixes the broken tests in bacon so LGTM

scripts/build_and_test.sh
```diff
@@ -15,9 +15,13 @@ BUILD_AND_TEST_TEST_SUITE_ID=771f3f55-7db9-11e5-80ff-22000b5181e9
 start_test_suite ${BUILD_AND_TEST_TEST_SUITE_ID}
  
 function build_and_test() {
-  ${MAVEN} clean deploy -s "${MAVEN_SETTINGS}"
+    sudo npm config set strict-ssl false
```
bdemers: Why `sudo` ? We _should_ be able to do a local install of these deps right?

scripts/build_and_test.sh
```diff
@@ -15,9 +15,13 @@ BUILD_AND_TEST_TEST_SUITE_ID=771f3f55-7db9-11e5-80ff-22000b5181e9
 start_test_suite ${BUILD_AND_TEST_TEST_SUITE_ID}
  
 function build_and_test() {
-  ${MAVEN} clean deploy -s "${MAVEN_SETTINGS}"
+    npm config set strict-ssl false
+    npm install
+    npm run start:mock-okta
+    ${MAVEN} clean package -s "${MAVEN_SETTINGS}"
```
bdemers: i'm guessing this should still be `deploy`

nikolaystoilov-okta: If we attempt to deploy we will try to sign and fail.

bdemers:
I'm +1 after the maven goal is fixed.

As an aside, I do think we should be able to run the tests in a language natural way.  And tests should be able to bootstrap themselves (with all of the fixtures they need), but that is a topic for another day.  If this fixes short term issues with CI, that is good (but we shouldn't be adding more `npm` to the build)

scripts/build_and_test.sh
bdemers: But CI will be running this script right ? which will need to sign/deploy the build.

Locally you should only need to `install` at most.  I'm not sure how CI is setup, but we may need one script for CI and one for dev (the or a flag in the build_and_test.sh script to not deploy)

tthompson-okta:
@rchild-okta you might want to give this a look.

mraible:
We might be able to use the [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin) to do the npm stuff. It also supports Yarn, which can speed things up quite a bit.

rchild-okta:
@mraible yep, @wdawson-okta has an example of this in the [samples-java-dropwizard](https://github.com/okta/samples-java-dropwizard/blob/master/pom.xml#L20) repo.

pom.xml
```diff
@@ -112,6 +113,53 @@
...
+                        <phase>initialize</phase>
+                    </execution>
+                    <execution>
+                        <id>npm configure</id>
+                        <goals>
+                            <goal>npm</goal>
+                        </goals>
+                        <phase>initialize</phase>
+                        <configuration>
+                            <arguments>config set strict-ssl false</arguments>
```
rchild-okta: Why do we need to set this to false?

pom.xml
```diff
@@ -112,6 +113,53 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
```
rchild-okta: Nice!

pom.xml
```diff
@@ -112,6 +113,53 @@
...
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>start mock server</id>
+                        <goals>
+                            <goal>npm</goal>
+                        </goals>
+                        <phase>generate-resources</phase>
+                        <configuration>
+                            <arguments>run pretest</arguments>
```
rchild-okta: @lboyette-okta, can you review this? Is it doing everything it needs to do (i.e. I'm guessing we need to do whatever is happening in posttest as well)

pom.xml
nikolaystoilov-okta: Without this we will not able to download the dependancies - unable to get local issuer certificate.

rchild-okta: That shouldn't generally be the case (we had seen that with an older version of node, but 5.6.0 should be clean). Where do you see that?

nikolaystoilov-okta: Here is the bacon run when I removed the section with the ssl settings:
http://bacon.aue1e.internal/#!/commits/oktasdk-java/sha/661185aefa4c7f39fbb47b84c1fdf92dbb82bf65

pom.xml
lboyette-okta: Yeah, `posttest` is never being run, so we're never asserting that the tests passed (all the endpoints were hit).

@nikolaystoilov-okta We need a way to run `npm run posttest` after the junit tests have run. `npm run posttest` will return a `0` on success or `1` on failure and output the failure to the command prompt.

scripts/build_and_test.sh
```diff
@@ -15,9 +15,9 @@ BUILD_AND_TEST_TEST_SUITE_ID=771f3f55-7db9-11e5-80ff-22000b5181e9
 start_test_suite ${BUILD_AND_TEST_TEST_SUITE_ID}
  
 function build_and_test() {
-  ${MAVEN} clean deploy -s "${MAVEN_SETTINGS}"
+    ${MAVEN} clean package -s "${MAVEN_SETTINGS}"
```
lboyette-okta: nit: let's keep the indentation consistent with the rest of the file (2 spaces)

pom.xml
```diff
@@ -112,6 +113,53 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>com.github.eirslett</groupId>
+                <artifactId>frontend-maven-plugin</artifactId>
+                <version>${frontend-maven-plugin.version}</version>
```
jmaldonado-okta: nit: A good practice is to define a `<dependencyManagement>` in the parent pom to manage all dependencies in your maven project from one place and avoid different modules using different versions indadvertedly.

See for example the following depednency in a sub-module: https://github.com/okta/okta-core/blob/master/server/framework/pom.xml#L22

Where the definition comes from the parent: https://github.com/okta/okta-core/blob/master/pom.xml#L404

The way you are defining versions will work for now but, you might hit wall ahead on the road and pull your hair once your project scales.

pom.xml
```diff
@@ -112,6 +113,53 @@
...
+                <configuration>
+                    <nodeVersion>v5.6.0</nodeVersion>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>install node and npm</id>
```
jmaldonado-okta: nit: as suggestion, try to avoid spaces in the id definitions. Some implementations do not support this approach.

pom.xml
jmaldonado-okta: I always consider disabling ssl as a hack :)

The error:
`[Thread-2] [ERROR] npm ERR! fetch failed https://registry.npmjs.org/asn1/-/asn1-0.2.3.tgz
[Thread-2] [WARNING] npm WARN retry will retry, error on last attempt: Error: unable to get local issuer certificate`

CI machines are configured to pull only from Artifactory using our Okta NPM registry `registry=https://artifacts.aue1d.saasure.com/artifactory/api/npm/npm-okta-master`. If I recall correctly, the npmrc also includes the path to our certificate through `cafile=FOO`.

But you are not pulling from Artifactory but `https://registry.npmjs.org` which definitely does not match or cafile. Also, you are using a maven plugin so I bet https://registry.npmjs.org is not trusted by our maven installations in CI. Since this is open/public, using Artifactory will not be an option.

So, you my suggestion is:
1. This should be reproducible locally. Once you do, trust the https://registry.npmjs.org cert in the maven local cert store and for your global npm installation. Once you succeed in building this locally, then:
2. Request Eng Productivity to repeat for CI machines.

scripts/build_and_test.sh
jmaldonado-okta: nit: Is this script used by external contributors as well? I wonder if we should start adding a differentiation between internal and external builds where they share the same core but things like `start_test_suite ${BUILD_AND_TEST_TEST_SUITE_ID}` should only be in the internal CI script.

pom.xml
nikolaystoilov-okta: updated

scripts/build_and_test.sh
nikolaystoilov-okta: fixed the spaces.
@jmaldonado-okta I am trying to do the minimum change to have the CI working again, so that I can commit my code change. If you want I can create separate jira to address the your dependancies and split script suggestions.

bdemers:
Is there an environment variable that we can set so the builds do not need to worry about this? This seems like a build env issue

nikolaystoilov-okta:
Will not merge this change.

PR #72, opened 2017-05-23
Authentication using activation token

nikolaystoilov-okta:
Added authentication method using activation token.
Extended authentication response with authentication token.

Resolves: OKTA-118829

This code was reviewed in the PR:  https://github.com/okta/okta-sdk-java/pull/64


bdemers:
+1, but we do NOT have authn support for the next version of the SDK yet

PR #108, opened 2017-07-07
IDEA: Try out using sdkman to manage java and maven versions

bdemers:
Not sure if this will work, so pushing a branch and testing out a travis build


bdemers:
👎 sdkman uses openjdk

PR #124, opened 2017-11-06
When looking for the maven 'target' directory use the last one

bdemers:
We cannot make any assumptions about the the folders above the 'target' directory, only below
NOTE: the release plugin creates a target/checkout/imp/target directory


bdemers:
Just disabled concurrent builds via travis and kicked it again

PR #136, opened 2017-12-11
handfull of minor fixes

bdemers:


bdemers:
NOTE: the Double property is needed for the upcoming Logs API

Issue #113, opened 2017-07-26
Invalid reference links provided in documentation


jmelberg-okta:
The site [docs.okta.com](http://docs.okta.com/) is referenced in `AbstractClient.java`, `AuthenticationScheme.java`, `ClientBuilder.java`, and `ResourceException.java`.

These should either be updated to point to the correct link, or removed.

PR #138, opened 2017-12-11
Remove invalid docs.okta.com links

bdemers:
Fixes: #113

### DIFF
--- a/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
+++ b/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
@@ -288,21 +288,17 @@ public interface ClientBuilder {
 
     /**
      * Overrides the default (very secure)
-     * <a href="http://docs.okta.com/rest/product-guide/#authentication-digest">Okta SSWS Digest
+     * <a href="https://developer.okta.com/docs/api/getting_started/design_principles.html#authentication">Okta SSWS Digest
      * Authentication Scheme</a> used to authenticate every request sent to the Okta API server.
      *
      * <p>It is not recommended that you override this setting <em>unless</em> your application is deployed in an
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/okta/sdk/client/ClientBuilder.java
```diff
@@ -288,21 +288,17 @@
 
     /**
      * Overrides the default (very secure)
-     * <a href="http://docs.okta.com/rest/product-guide/#authentication-digest">Okta SSWS Digest
+     * <a href="https://developer.okta.com/docs/api/getting_started/design_principles.html#authentication">Okta SSWS Digest
```
jmelberg-okta: Can you remove the `.html` here? Makes the URL look much better. 😄

api/src/main/java/com/okta/sdk/resource/ResourceException.java
```diff
@@ -77,8 +77,9 @@ public int getStatus() {
     }
 
     /**
-     * Get the Okta Error Code
-     * Check http://docs.okta.com/errors/ for the list of Okta Error Codes
+     * Get the Okta Error Code, <a href="https://developer.okta.com/reference/error_codes/index">click here</a> for the
```
jmelberg-okta: This page will not render when we move 100% over to CloudFront.

Can you update this to: `/reference/error_codes/`

bdemers: I'm pretty sure I pulled that from a google search, so that might be a bigger issue (but i can update it)

jmelberg-okta: **FYI:** With [this change](https://github.com/okta/okta.github.io/pull/1550) we are updating the `sitemap` to use pretty-URL format: `https://site.com/page` instead of `https://site.com/page.html`.

PR #143, opened 2017-12-12
Expose property util methods to any generated object marked with 'x-okta-extensible' i.e. UserProfile

bdemers:
This adds the following methods to mimic what Okta supports with custom profile schemas:

    String getString(String key);
    Double getNumber(String key);
    Boolean getBoolean(String key);
    Integer getInteger(String key);
    List<String> getStringList(String key);
    List<Double> getNumberList(String key);
    List<Integer> getIntegerList(String key);


bdemers:
Tests already exist, this just basically marks the methods as public

PR #144, opened 2017-12-13
IDEA: move config parsing from spring module into the api module for reuse (DO NOT MERGE)

bdemers:
Was thinking about how we could support Spring, Guice, and CDI with some common code.

Basically move the logic of parsing a configuration object into the core code.
The way to pull in configuration differs between DI
- Spring - you can it is a trivial annotation
- with Guice and CDI, there are many different ways to do this and no real standard
- CDI has NO way of doing this with out an extension, and it isn't the job of this SDK to force dependencies on someone

Each implementation allows for the optional injection of a CacheManager and ClientConfiguration

Few warts are in here, but should be good enough for discussion
- the ClientConfiguration classname conflicts with the implementation in the DefaultClientBuilder
- ClientConfiguration's nested structure comes from the way things work with Spring @ConfigurationProperties works


bdemers:
@mraible @dogeared I wanted to hack on this to see if there _was_ any common code here, and what that would look like. Though I'm not sure if I'd want to take on supporting Guice and CDI modules (if there isn't many people asking for this)

mraible:
@bdemers Guice could be useful for Android. Especially if it works with [Dagger](http://square.github.io/dagger/) which is based on Guide AFAIK. Not sure about CDI, but I say +1 if it's easy to create. Maintaining should be easy with the way specs move for Java EE. That could change at Eclipse though.

bdemers:
Closing for now

Issue #146, opened 2018-01-17
Cannot get resetPasswordUrl from user.forgotPassword


me717:
The ForgotPasswordResponse object returned by forgotPassword() and forgotPassword(sendEmail, userCredentials) method of the user class never has the resetPasswordUrl property initialized.

This appears to be because the backend api only return that field if sendEmail is false, but no user credentials are included in the request (as described [here](https://developer.okta.com/docs/api/resources/users.html#forgot-password)).  This is a call that is presently impossible to make from this sdk.  @bdemers Would it be possible to get a forgotPassword(sendEmail) method, or be able to include null credentials?

bdemers:
For a work around try calling `user.forgotPassword(false, client.instantiate(UserCredentials.class))` this should make the request with an empty payload.

me717:
That results in an error with the following causes: `password: The field cannot be left blank;recoveryQuestion: The field cannot be left blank`

bdemers:
@me717 OK we will look into that.  Thanks for trying it out and getting back to us!

PR #147, opened 2018-01-18
Allow for optional bodies.

bdemers:
`User.forgotPassword()` has an optional arg `UserCredentials`.  If this argument can now be null.
Included IT

Fixes: #146

### DIFF
--- a/swagger-templates/src/main/resources/OktaJavaImpl/apiMethodPostResource.mustache
+++ b/swagger-templates/src/main/resources/OktaJavaImpl/apiMethodPostResource.mustache
@@ -27,7 +27,12 @@
         {{/vendorExtensions.bodyIsSelf}}
         {{^vendorExtensions.bodyIsSelf}}
             {{#bodyParam}}
+                {{#bodyParam.required}}
             {{bodyParam.paramName}},
+                {{/bodyParam.required}}
+                {{^bodyParam.required}}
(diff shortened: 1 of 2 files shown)


bdemers:
**NOTE:** `User.forgotPassword` is the only method that has an optional body parameter

swagger-templates/src/main/resources/OktaJavaImpl/apiMethodPostResource.mustache
```diff
@@ -27,7 +27,12 @@
         {{/vendorExtensions.bodyIsSelf}}
         {{^vendorExtensions.bodyIsSelf}}
             {{#bodyParam}}
+                {{#bodyParam.required}}
             {{bodyParam.paramName}},
+                {{/bodyParam.required}}
```
jmelberg-okta: The indentation makes this a little difficult to read, but I'm assuming it is necessary because it'll throw off the rendered template, correct?

If not, can we make these line-up so it is easier to read?:
```
{{#bodyParam}}
  {{#bodyParam.required}}
  {{bodyParam.paramName}},
  {{/bodyParam.required}}
...
```

bdemers: Correct, it has been bugging me too, I don't have a good solution for it

PR #151, opened 2018-01-26
Move HREF resolution out into a new class ResourceHrefResolver

bdemers:
Added implementations for HAL and static resolution based on the current class and properties (and tests)

### DIFF
--- a/impl/src/main/java/com/okta/sdk/impl/resource/AbstractResource.java
+++ b/impl/src/main/java/com/okta/sdk/impl/resource/AbstractResource.java
@@ -21,8 +21,6 @@ import com.okta.sdk.lang.Collections;
 import com.okta.sdk.lang.Strings;
 import com.okta.sdk.resource.CollectionResource;
 import com.okta.sdk.resource.Resource;
-import com.okta.sdk.resource.application.AppUser;
-import com.okta.sdk.resource.application.Application;
 
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -54,6 +52,7 @@ public abstract class AbstractResource extends AbstractPropertyRetriever impleme
     private String href = null;
     protected volatile boolean dirty;
     private volatile boolean materialized;
+    private final ResourceHrefResolver hrefResolver;
 
     protected AbstractResource(InternalDataStore dataStore) {
         this(dataStore, null);
@@ -65,6 +64,7 @@ public abstract class AbstractResource extends AbstractPropertyRetriever impleme
         this.dirtyProperties = new LinkedHashMap<>();
         this.deletedPropertyNames = new HashSet<>();
         this.properties = new LinkedHashMap<>();
+        this.hrefResolver = new OktaResourceHrefResolver();
         setInternalProperties(properties);
     }
 
@@ -102,7 +102,7 @@ public abstract class AbstractResource extends AbstractPropertyRetriever impleme
                 } else {
                     this.properties = properties;
                 }
-                setResourceHref(getSelfHref(properties));
+                setResourceHref(hrefResolver.resolveHref(properties, getClass()));
                 // Don't consider this resource materialized if it is only a reference.  A reference is any object that
                 // has only one 'href' property.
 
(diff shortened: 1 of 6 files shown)


bdemers:
(Part 2 of breaking up cacheing PR)

impl/src/main/java/com/okta/sdk/impl/resource/OktaResourceHrefResolver.java
```diff
@@ -0,0 +1,91 @@
...
+            }
+            if (Application.class.isAssignableFrom(clazz)) {
+                Map<String, ?> self = getMapValue(links, "users");
+                if (!Collections.isEmpty(self)) {
+                    String href = self.get("href").toString();
+                    return href.substring(0, href.lastIndexOf("/users"));
+                }
+            }
+        }
+
```
vijetmahabaleshwar-okta: So we're fixing the self href for AppUsers, Application, Group and GroupRule objects right? If so, can you add more comments about the same?

impl/src/main/java/com/okta/sdk/impl/resource/AbstractResource.java
```diff
@@ -102,7 +102,7 @@ public void setInternalProperties(Map<String, Object> properties) {
                 } else {
                     this.properties = properties;
                 }
-                setResourceHref(getSelfHref(properties));
+                setResourceHref(hrefResolver.resolveHref(properties, getClass()));
```
vijetmahabaleshwar-okta: `resolveHref` can return `null` right? Do we need to handle the case when its `null`

impl/src/main/java/com/okta/sdk/impl/resource/OktaResourceHrefResolver.java
bdemers: These resources are not going to change in the short term.

impl/src/main/java/com/okta/sdk/impl/resource/AbstractResource.java
bdemers: Nope `hrefResolver` is set in the constructor and is final

bdemers: sorry, i read that comment wrong.
Yes, it can return `null`, but that fine. For example in the case of a password reset body there would be no HREF (there is no way to request that resource again)

impl/src/main/java/com/okta/sdk/impl/resource/OktaResourceHrefResolver.java
```diff
@@ -0,0 +1,91 @@
...
+ * </ul>
+ *
+ * @since 0.11.0
+ */
+public class OktaResourceHrefResolver implements ResourceHrefResolver {
+
+    private final ResourceHrefResolver halResourceHrefResolver = new HalResourceHrefResolver();
+
+    @Override
+    public <R extends Resource> String resolveHref(Map<String, ?> properties, Class<R> clazz) {
```
vijetmahabaleshwar-okta: Just a thought - Since we only return "self" href link, should this method be named `resolveSelfHref`? Will we ever need  "non-self" href links in the future?

impl/src/test/groovy/com/okta/sdk/impl/resource/HalResourceHrefResolverTest.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    foo: "bar"
+            ]]]
+
+        assertThat resolver.resolveHref(Collections.emptyMap(), User), nullValue()
+        assertThat resolver.resolveHref(nullHrefProps, User), nullValue()
+        assertThat resolver.resolveHref(missingHrefProps, User), nullValue()
+    }
+
+    @Test
+    void resourceWithSelfLinkTest() {
```
vijetmahabaleshwar-okta: Do we need to add test with multiple href links, not just self href?

impl/src/main/java/com/okta/sdk/impl/resource/OktaResourceHrefResolver.java
bdemers: I think the `self` link is an implementation detail, in Okta's case we use HAL style links.  At Stormpath we used a root level property of `href`.

PR #155, opened 2018-01-31
Use Input/Output Streams instead of String in MapMarshaller

bdemers:


impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -473,7 +479,7 @@ private Response execute(Request request) throws ResourceException {
         Map<String, Object> out = null;
 
         if (response.hasBody()) {
-            out = mapMarshaller.unmarshall(response.getBody(), response.getHeaders().getLinkMap());
+            out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
```
jmelberg-okta: Doesn't `unmarshal()` return `void`? See [L27 of `impl/src/main/java/com/okta/sdk/impl/ds/MapMarshaller.java`](https://github.com/okta/okta-sdk-java/pull/155/files#diff-1641488eafa261d2df9a52b0882796bbR27)

bdemers: `unmarshal` returns a map, `marshal` now returns `void`

PR #156, opened 2018-01-31
Javadoc cleanup

bdemers:
Fixed a bunch of issues detected by Intellij's Inspection


api/src/main/java/com/okta/sdk/resource/ResourceException.java
```diff
@@ -35,9 +35,9 @@
     /**
      * Ensures the message used for the exception (i.e. exception.getMessage()) reports the {@code developerMessage}
      * returned by the Okta API Server.  The regular Okta response body {@code message} field is targeted
-     * at applicadtion end-users that could very likely be non-technical.  Since an exception should be helpful to
+     * at application end-users that could very likely be non-technical.  Since an exception should be helpful to
      * developers, it is better to show a more technical message.
-     * <p/>
+     * <p>
```
jmelberg-okta: Looks like we accidentally removed: `</p>`

api/src/main/java/com/okta/sdk/resource/ResourceException.java
```diff
@@ -103,18 +103,14 @@ public String getId() {
...
     /**
      * Returns the underlying REST {@link Error} returned from the Okta API server.
-     * <p/>
+     * <p>
      * Because this class's {@link #getMessage() getMessage()} value returns a developer-friendly message to help you
      * debug when you see stack traces, you might want to acquire the underlying {@code Error} to show an end-user
      * the simpler end-user appropriate error message.  The end-user error message is non-technical in nature - as a
      * convenience, you can show this message directly to your application end-users.
-     * <p/>
+     * <p>
```
jmelberg-okta: Same comment as above.

api/src/main/java/com/okta/sdk/resource/ResourceException.java
bdemers: Intentional, it fails to lint this way (you cannot have self closing tags):
https://stackoverflow.com/questions/33787753/javadoc-closing-tags

PR #157, opened 2018-01-31
Remove dead code

bdemers:


bdemers:
Agreed, this heavily depend on refactor tooling and test coverage

PR #159, opened 2018-01-31
Simplify DefaultDataStore anonymous classes with lambdas

bdemers:
(mostly automated refactoring)

### DIFF
--- a/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
+++ b/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
@@ -511,20 +511,7 @@ public class DefaultDataStore implements InternalDataStore {
         if (href == null || href.length() < 5) {
             return false;
         }
-
-        char c = href.charAt(0);
-        if (c == 'h' || c == 'H') {
-            c = href.charAt(1);
-            if (c == 't' || c == 'T') {
-                c = href.charAt(2);
-                if (c == 't' || c == 'T') {
-                    c = href.charAt(3);
-                    return c == 'p' || c == 'P';
-                }
-            }
-        }
-
-        return false;
+        return href.regionMatches(true, 0, "http", 0, 4);
     }
 
     protected String qualify(String href) {
--- a/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
+++ b/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
@@ -177,7 +177,7 @@ public class DefaultDataStore implements InternalDataStore {
     @SuppressWarnings("unchecked")
     @Override
     public <T extends Resource> T getResource(String href, Class<T> clazz) {
-        return getResource(href, clazz, (Map<String, Object>) null);
+        return getResource(href, clazz, null);
     }
 
     public <T extends Resource> T getResource(String href, Class<T> clazz, Map<String, Object> queryParameters) {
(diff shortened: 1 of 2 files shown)


impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -278,55 +275,52 @@ private ResourceAction getPostAction(ResourceDataRequest request, Response respo
...
-                }
+            Request request = new DefaultRequest(method, href1, qs1, httpHeaders, body, length);
 
-                ResourceAction responseAction = getPostAction(req, response);
+            Response response = execute(request);
+            Map<String, Object> responseBody = getBody(response);
 
-                return new DefaultResourceDataResult(responseAction, uri, returnType, responseBody);
+            if (Collections.isEmpty(responseBody)) {
+                // Fix for https://github.com/stormpath/stormpath-sdk-java/issues/218
```
jmelberg-okta: Since this is licensed to Stormpath, we need to update the header to denote that they were changed.

See the [LICENSE](https://github.com/okta/okta-sdk-java/blob/5505055ee0974adf6f85b0d0efbe3fd5ad308401/LICENSE#L91).

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -529,9 +519,7 @@ protected boolean isFullyQualified(String href) {
                 c = href.charAt(2);
                 if (c == 't' || c == 'T') {
```
jmelberg-okta: I'd recommend cleaning this up by using `toLowerCase`

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
bdemers: see: https://github.com/okta/okta-sdk-java/blob/master/THIRD-PARTY-NOTICES#L116

jmelberg-okta: IIRC including a notice to the `THIRD-PARTY-NOTICES` file doesn't indicate that you've modified the contents of the file.

@robertjd Do you recall what Legal requires us to do?

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
bdemers: I've been eying this one too, I'm not sure why this was originally done this way.  I'm going to follow up with another PR for things like this.

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
bdemers: ;) https://github.com/okta/okta-sdk-java/blob/master/THIRD-PARTY-NOTICES#L105

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -521,22 +511,7 @@ protected boolean isFullyQualified(String href) {
...
-                    c = href.charAt(3);
-                    if (c == 'p' || c == 'P') {
-                        return true;
-                    }
-                }
-            }
-        }
-
-        return false;
+        return href.regionMatches(true, 0, "http", 0, 4);
```
bdemers: @jmelberg-okta ^

jmelberg-okta:
With these kind of changes, @vijetmahabaleshwar-okta should probably take a look.

Still need to know if adding that line to our `THIRD_PARTY_NOTICES` file is sufficient before merging though.

bdemers:
Re the notice, it is nothing new. 90% of this code is from the Stormpath Java SDK ;)

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
jmelberg-okta: As mentioned, we should include the following on every modified file:
> Modifications Copyright (C) 2018 Okta, Inc. and/or its affiliates.

Lets pick this up that change on another PR

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -278,55 +275,52 @@ private ResourceAction getPostAction(ResourceDataRequest request, Response respo
...
 
-        FilterChain chain = new DefaultFilterChain(this.filters, new FilterChain() {
-            @Override
-            public ResourceDataResult filter(final ResourceDataRequest req) {
+        FilterChain chain = new DefaultFilterChain(this.filters, req -> {
 
-                CanonicalUri uri = req.getUri();
-                String href = uri.getAbsolutePath();
-                QueryString qs = uri.getQuery();
+            CanonicalUri uri1 = req.getUri();
```
vijetmahabaleshwar-okta: So, you can't reuse variable names with 2 lambda expressions in the same method?

bdemers: It should be possible, it would be scoped different, it's just not clear when they are reused

vijetmahabaleshwar-okta: In that case, I'd have preferred the first name to be `uri1` and the 2nd to be `uri2`. Just a suggestion. Approved anyway :)

PR #162, opened 2018-02-02
Add support for swagger password format

bdemers:
Update UserBuilder and UserIT tests to use char[] instead of string


src/swagger/api.yaml
```diff
@@ -29,7 +29,7 @@
       "name": "Apache-2.0",
       "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
     },
-    "version": "0.11.0"
+    "version": "0.12.0"
```
bdemers: This file may change slightly once we have an official swagger spec version, but I'm only expecting the version to change at this point

bdemers: No diffs with the 0.12.0 release, so this files does not need to change

impl/src/test/java/com/okta/sdk/impl/resource/StubResource.java
```diff
@@ -31,8 +31,10 @@
     public final  MapProperty mapProp = new MapProperty("mapPropKey");
     public final  ResourceListProperty<NestedStubResource> resourceListProp = new ResourceListProperty<>("resourceListPropKey", NestedStubResource.class);
     public final  StringProperty stringProp = new StringProperty("stringPropKey");
+    public final  CharacterArrayProperty charArrayProp1 = new CharacterArrayProperty("charArrayPropKey1");
+    public final  CharacterArrayProperty charArrayProp2 = new CharacterArrayProperty("charArrayPropKey2");
 
-    private final Map<String, Property> propertyDescriptors = AbstractResource.createPropertyDescriptorMap(booleanProp, dateProp, enumListProp, enumProp, integerProp, doubleProp, listProp, mapProp, resourceListProp, stringProp);
+    private final Map<String, Property> propertyDescriptors = AbstractResource.createPropertyDescriptorMap(booleanProp, dateProp, enumListProp, enumProp, integerProp, doubleProp, listProp, mapProp, resourceListProp, stringProp, charArrayProp1, charArrayProp2);
```
jmelberg-okta: nit: Are there common lint rules you could impose to break this into a multiline assignment?

bdemers: Lots of standards

bdemers: I like Python's PEP8, and PHP has something similar, Java has a few common code styles.

PR #165, opened 2018-02-02
Reduced scope of a bunch of methods and classes

bdemers:
Other minor style fixes from introspect code in Intellij


api/src/main/java/com/okta/sdk/http/HttpRequests.java
```diff
@@ -47,7 +47,7 @@
      * @return a new HttpRequestBuilder that can be used to construct a new {@link HttpRequest} instance.
      * @throws IllegalArgumentException if the method argument is {@code null}.
```
jmelberg-okta: This comment needs to be updated.

impl/src/main/java/com/okta/sdk/impl/cache/DefaultCache.java
```diff
@@ -126,7 +126,7 @@ public DefaultCache(String name, Map<K, Entry<V>> backingMap) {
      * @see #setTimeToIdle(com.okta.sdk.lang.Duration)
      * @see #setTimeToLive(com.okta.sdk.lang.Duration)
      */
-    public DefaultCache(String name, Map<K, Entry<V>> backingMap, Duration timeToLive, Duration timeToIdle) throws IllegalArgumentException {
+    public DefaultCache(String name, Map<K, Entry<V>> backingMap, Duration timeToLive, Duration timeToIdle) {
```
jmelberg-okta: Same comment as above.

impl/src/main/java/com/okta/sdk/impl/http/support/Version.java
```diff
@@ -28,39 +28,39 @@
 public class Version {
 
     private static final String CLIENT_VERSION = lookupClientVersion();
+    private static final String VERSION_FILE = "/com/okta/sdk/version.properties";
 
     public static String getClientVersion() {
         return CLIENT_VERSION;
     }
 
     private static String lookupClientVersion() {
```
jmelberg-okta: Shouldn't this include `throws RuntimeException`?

bdemers: Nope, `RuntimeExceptions` are not declared they are not a checked exception

api/src/main/java/com/okta/sdk/http/HttpRequests.java
bdemers: This doc line is still actually valid. The exception is thrown when `method` is null. It's just not a checked exception

impl/src/main/java/com/okta/sdk/impl/cache/DefaultCache.java
bdemers: Same as above `Assert.notNull` does throw a `IllegalArgumentException` it just doesn't belong in the method signature because it is a RuntimeException

PR #174, opened 2018-04-05
Prep 1.0.1 release

bdemers:
lumping a few small commits together in one PR, you can review them independently or all at once
The first 2 are pretty isolated, the last three are CI build cleanup


.editorconfig
```diff
@@ -0,0 +1,75 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+[*]
+charset=utf-8
+end_of_line=lf
+insert_final_newline=false
```
vijetmahabaleshwar-okta: Have we decided that `insert_final_newline` should be false? I thought we needed it. 
Either way, not a big deal.

src/ci/release.sh
```diff
@@ -26,7 +26,7 @@ TAG_NAME="okta-sdk-root-${NEW_VERSION}" # default release plugin tag format
 
 # TODO: we must use our local maven settings file as this script is NOT ready for triggered by travis
```
vijetmahabaleshwar-okta: Is this TODO related to the change you made? i.e. does this still apply?

mvnw
```diff
@@ -0,0 +1,286 @@
...
+    ls=`ls -ld "$PRG"`
+    link=`expr "$ls" : '.*-> \(.*\)$'`
+    if expr "$link" : '/.*' > /dev/null; then
+      PRG="$link"
+    else
+      PRG="`dirname "$PRG"`/$link"
+    fi
+  done
+
+  saveddir=`pwd`
```
vijetmahabaleshwar-okta: would `pushd`, `popd` work here?

.editorconfig
bdemers: No official resolution yet

src/ci/release.sh
bdemers: Still applies, still a little work to do on that front

mvnw
bdemers: all of the `mvnw` is generated from: https://github.com/takari/maven-wrapper

Issue #101, opened 2017-06-23
Support for Identity Providers API


AlainODea:
We use the [Identity Providers API](https://developer.okta.com/docs/api/resources/idps.html) in production to:

 1. Allow customers to create and update their IdPs with necessary security restrictions from our application for SSO self-enrolment.
 2. Allow customers to see logs specific to their IdP for SSO self-troubleshooting (used to get target ID for use with Events API)

AlainODea:
@bdemers this is the IdP API we are using in production. We can't migrate to the new style of SDK until this is in it.

AlainODea:
@bdemers when will the Identity Providers API be added to this SDK?

bdemers:
Hey @AlainODea sorry I didn't reply to this one! or #102. 

I'll pass the buck to @nbarbettini ;)

AlainODea:
No worries 😄

I think Identity Providers can be supported (in a basic way) with changes to api.yaml and running the Swagger codegen. If I understand it correctly, most of API client plumbing (per API endpoint) is generated by Swagger. That's awesome!

nbarbettini:
@AlainODea The API client plumbing is indeed generated via Swagger (OpenAPI). The IdP API isn't one we have worked on adding yet, but it's in the backlog. 👍

nbarbettini:
@AlainODea Just curious, other than this API, what other features would you like to see in this SDK? What would make it a really great Java experience?

AlainODea:
I work on an Cloud-based multitenant app that integrates extensively with Okta.

Okta enables SAML SSO, MFA, and seamless access to a community/support site for the tenants and their users.

The major things we use are:
* Users: so we can set passwords, unlock, and reset password
* UserGroups: to move Users to groups that have an MFA enforcement policy and to lock Inbound SAML users to the correct tenant isolation config
* AppUsers: to individually assign users to federated apps like the community/support site (some customers prefer to restrict this to managers or IT on their side)
* AppUserGroups: to assign entire tenants and all their users to our Okta Sign-On Widget app and our Inbound SAML app
* Identity Providers: (this) to support SSO self-enrolment
* Factors: to support MFA self-enrolment
* Events (or equivalent): to get Identity Provider logs for self-enrolment and self-troubleshooting by our tenants
* (emerging req) Schemas: to add custom attributes to support user roles on Inbound SAML 
[...]

PR #181, opened 2018-04-20
:seedling: Add Identity Providers API

AlainODea:
You're probably not accepting PRs to legacy. I'm sharing this in case it may be useful to @nbarbettini 😃 

It adds:

 - Support for Identity Providers

 - Support for Keys

 - Support for Metadata

This includes significant contributions by @chris-whittick. I rebased it on legacy and squashed to a single commit to keep it simple.


AlainODea:
Related to #101.

bdemers:
@AlainODea, you are right we won't be adding new features to the legacy branch. But sharing code is always a good thing, and helps us see your use cases!

bdemers:
Opened #199 to track this.  Thanks again!

PR #182, opened 2018-04-23
Expose for authn

bdemers:
Allows core components to be reused by other Okta SDKs
**Depends on:** 
- https://github.com/okta/openapi/pull/119


THIRD-PARTY-NOTICES
```diff
@@ -13,93 +13,42 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 This project includes:
```
bdemers: Update to parent pom, fixes the generated notice, to ONLY include compile dependencies

README.md
```diff
@@ -148,3 +148,9 @@ for (User tmpUser : users) {
 ## Contribute to the Project
 
 Take a look at the (contribution guide)[CONTRIBUTING.md] and the [build instructions wiki](https://github.com/okta/okta-sdk-java/wiki/Build-It) (though just cloning the repo and running `mvn install` should get you going).
+
+## Versioning
```
bdemers: @dogeared @mraible @laura-rodriguez thoughts on this topic?
I don't think this is likely to effect anyone, but I've added a couple non-`default` methods to an interface in this commit, so I feel like it needs to be stated someplace.

nbarbettini: nit: affect not effect

laura-rodriguez: it looks like @nbarbettini is OK with this, so I am :)

api/src/main/java/com/okta/sdk/lang/Locales.java
```diff
@@ -0,0 +1,202 @@
...
+        final int sz = cs.length();
+        for (int i = 0; i < sz; i++) {
+            if (!Character.isUpperCase(cs.charAt(i))) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static boolean isNumeric(final CharSequence cs) {
```
jmelberg-okta: It might make more sense to pull these methods out into a generic Util file and live under `/imp/util`.

impl/src/main/java/com/okta/sdk/impl/http/support/SdkUserAgentProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.okta.sdk.http.UserAgentProvider;
+
+/**
+ * @since 1.1.0
+ */
+public class SdkUserAgentProvider implements UserAgentProvider {
+
+    private static final String OKTA_SDK_STRING = "okta-sdk-java";
+    private static final String VERSION_SEPARATOR = "/";
```
jmelberg-okta: Since we're concatenating `OKTA_SDK_STRING` and `VERSION_SEPARATOR` below, why don't we make this as one variable?

src/japicmp/postAnalysisScript.groovy
```diff
@@ -0,0 +1,47 @@
...
+
+        def methodIt = jApiClass.getMethods().iterator()
+        while (methodIt.hasNext()) {
+            def method = methodIt.next()
+            def methodChanges = method.getCompatibilityChanges()
+            methodChanges.remove(METHOD_NEW_DEFAULT)
+
+            if (methodChanges) {
+                println("  method ${method.getName()} ${methodChanges}")
+
```
jmelberg-okta: nit: Extra line

api/src/main/java/com/okta/sdk/lang/Locales.java
bdemers: Not a bad idea, but I don't think they would be used outside this context (and i'd rather keep them private until needed)

impl/src/main/java/com/okta/sdk/impl/http/support/SdkUserAgentProvider.java
bdemers: +1, this likely shouldn't even be a constant

bdemers:
@jmelberg-okta thanks!
Take a look at the [Build-iT](https://github.com/okta/okta-sdk-java/wiki/Build-It) wiki page for build instructions. We may build an example in the authn repo, but not in this one (an example here would also create a circular dependency when developing)

PR #185, opened 2018-04-30
Remove hardcoded supported packages list.

bdemers:
This can now be extended via a new service `ResourceFactoryConfig`

### DIFF
--- a/impl/src/main/java/com/okta/sdk/impl/ds/DefaultResourceFactory.java
+++ b/impl/src/main/java/com/okta/sdk/impl/ds/DefaultResourceFactory.java
@@ -21,10 +21,15 @@ import com.okta.sdk.resource.Resource;
 
 import java.lang.reflect.Constructor;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static java.util.stream.StreamSupport.stream;
 
 /**
  * @since 0.5.0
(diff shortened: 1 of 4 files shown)


impl/src/main/java/com/okta/sdk/impl/ds/DefaultResourceFactory.java
```diff
@@ -33,7 +38,7 @@
 
     private final InternalDataStore dataStore;
 
-    private static final List<String> SUPPORTED_PACKAGES = Arrays.asList("com.okta.sdk.", "com.okta.authn.sdk.");
```
bdemers: @jmelberg-okta one of your comments on the previous review had me thinking.  There should not be any refs to `authn` in this repo if we can help it.

impl/src/main/java/com/okta/sdk/impl/ds/ResourceFactoryConfig.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Set;
+
+/**
+ * @since 1.1.0
+ */
+public interface ResourceFactoryConfig {
+
+    Set<String> getSupportedPackages();
+}
```
jmelberg-okta: You should add an `.editorconfig` file to detect missing or omit newlines. ;)

impl/src/main/java/com/okta/sdk/impl/ds/DefaultResourceFactory.java
```diff
@@ -160,12 +165,22 @@ public DefaultResourceFactory(InternalDataStore dataStore) {
...
         }
 
         return args.toArray();
     }
+
+    private static Set<String> getSupportedPackages() {
+        Set<String> result = new HashSet<>();
+        stream(ServiceLoader.load(ResourceFactoryConfig.class).spliterator(), false)
+                    .map(config -> config.getSupportedPackages().stream()
+                               .map(it -> it +".")
```
jmelberg-okta: `+".")` -> `+ ".")` (space)

impl/src/main/java/com/okta/sdk/impl/ds/ResourceFactoryConfig.java
bdemers: It has one ;) though maybe it isn't configure right? it has: `insert_final_newline=false`

PR #186, opened 2018-05-01
Added support to generate backward compatible method signatures when new optional query args are added

bdemers:
NOTE: initial support only allows for 2 versions (old and new), this may need to change
going forward i.e. if a method signature is updated twice.


swagger-templates/src/main/resources/OktaJavaImpl/queryPartialBackwardsCompat.mustache
```diff
@@ -0,0 +1,19 @@
+{{!
+    Copyright 2017 Okta
```
bretterer: Should this change to 2017-Present

bdemers:
OH! yeah, i should have done that.  Good call

PR #193, opened 2018-05-30
Add proxy settings example to README

bdemers:
Fixes: #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,15 @@ for (User tmpUser : users) {
 }
 ```
 
+## Proxy Configuration
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,6 +145,15 @@ for (User tmpUser : users) {
 }
 ```
 
+## Proxy Configuration
+
+Use the standard [Java proxy system properties](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) ([Spring documentation](https://docs.spring.io/spring-integration/reference/html/http.html#http-proxy)) to configure proxy support.
+
+```txt
+-Dhttps.proxyHost={your.proxt.host} \
```
jmelberg-okta: nit: Just to make this a bit more clear, lets add the proceeding start command.

```bash
java thisAwesomeSDK
-Dhttps.proxyHost={your.proxt.host} \
-Dhttps.proxyPort={your.proxy.port}
```

bdemers: Good point, but this is very dependent on the application, maybe this should be a table instead of a code block 🤔

PR #194, opened 2018-05-30
Generate xref source to make it easer to view the generated source

bdemers:
NOTE: CI will fail until the parent pom is released

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>com.okta</groupId>
         <artifactId>okta-parent</artifactId>
-        <version>4</version>
+        <version>5</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -20,7 +20,7 @@
     <parent>
         <groupId>com.okta</groupId>
         <artifactId>okta-parent</artifactId>
-        <version>4</version>
+        <version>5-SNAPSHOT</version>
```
jmelberg-okta: Looks like modifying the version is causing your tests to fail. Can you double check?

bdemers: that is the reason, the SNAPSHOT isn't publish, so this is just queued up until that parent is released

PR #197, opened 2018-06-15
Parse X-Rate-Limit-Reset header to calculate 429 retry delay

bdemers:
Replaces reflection loading of HttpClientRequestExecutor with a factory configured with ServiceLoader
Use `ClientConfiguration` as constructor arg to make method signature more stable (and allow for additional arguments in the future)
Added `okta.client.retry.maxElapsed` configuration

TODO: update configuration for 429 min/max randomization

### DIFF
--- a/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
+++ b/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
@@ -208,6 +208,7 @@ public interface ClientBuilder {
     String DEFAULT_CLIENT_PROXY_HOST_PROPERTY_NAME = "okta.client.proxy.host";
     String DEFAULT_CLIENT_PROXY_USERNAME_PROPERTY_NAME = "okta.client.proxy.username";
     String DEFAULT_CLIENT_PROXY_PASSWORD_PROPERTY_NAME = "okta.client.proxy.password";
+    String DEFAULT_CLIENT_RETRY_MAX_ELAPSED_PROPERTY_NAME = "okta.client.retry.maxElapsed";
 
     /**
      * Allows specifying an {@code ApiKey} instance directly instead of relying on the
@@ -328,6 +329,14 @@ public interface ClientBuilder {
      * @return the ClientBuilder instance for method chaining
      */
     ClientBuilder setOrgUrl(String baseUrl);
+
+    /**
+     * Sets the maximum number of milliseconds to wait when retrying before giving up.
+     *
+     * @param maxElapsed retry max elapsed duration in milliseconds
+     * @return the ClientBuilder instance for method chaining
+     */
+    ClientBuilder setRetryMaxElapsed(int maxElapsed);
 
     /**
      * Constructs a new {@link Client} instance based on the ClientBuilder's current configuration state.
--- a/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
+++ b/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
@@ -209,6 +209,7 @@ public interface ClientBuilder {
     String DEFAULT_CLIENT_PROXY_USERNAME_PROPERTY_NAME = "okta.client.proxy.username";
     String DEFAULT_CLIENT_PROXY_PASSWORD_PROPERTY_NAME = "okta.client.proxy.password";
     String DEFAULT_CLIENT_RETRY_MAX_ELAPSED_PROPERTY_NAME = "okta.client.retry.maxElapsed";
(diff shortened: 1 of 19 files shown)


api/src/main/java/com/okta/sdk/client/ClientBuilder.java
```diff
@@ -208,6 +208,8 @@
     String DEFAULT_CLIENT_PROXY_HOST_PROPERTY_NAME = "okta.client.proxy.host";
     String DEFAULT_CLIENT_PROXY_USERNAME_PROPERTY_NAME = "okta.client.proxy.username";
     String DEFAULT_CLIENT_PROXY_PASSWORD_PROPERTY_NAME = "okta.client.proxy.password";
+    String DEFAULT_CLIENT_RETRY_MAX_ELAPSED_PROPERTY_NAME = "okta.client.requestTimeout";
```
robertjd: Nit: the config var is now `requestTimeout`, but we are still using "max elapsed" in property names and method names

httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -81,7 +84,7 @@
     /**
      * Maximum exponential back-off time before retrying a request
      */
-    private static final int MAX_BACKOFF_IN_MILLISECONDS = 20 * 1000;
+    private static final int DEFAULT_MAX_BACKOFF_IN_MILLISECONDS = 20 * 1000;
 
     private static final int DEFAULT_MAX_RETRIES = 4;
```
robertjd: Should be 2 per our spec?

httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -342,26 +408,37 @@ private boolean isRedirect(HttpResponse response) {
...
-            delay = this.backoffStrategy.getDelayMillis(retries);
-        } else {
-            long scaleFactor = 300;
-            if (previousException != null && isThrottlingException(previousException)) {
-                scaleFactor = 500 + random.nextInt(100);
+            delay = Math.min(this.backoffStrategy.getDelayMillis(retries), timeElapsedLeft);
+        } else if (httpResponse != null && httpResponse.getStatusLine().getStatusCode() == 429) {
+            delay = get429DelayMillis(httpResponse);
+            if (!shouldRetry(retries, timeElapsed + delay)) {
+                throw new RestException("HTTP 429: Too Many Requests.  Exceeded request rate limit in the allotted amount of time.");
```
robertjd: Do we need to add this error text?  Can we not pass through what we get on the response?

httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
bdemers: The current implementation uses 4, for backwards compatibility

httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
bdemers: It does, via this: https://github.com/okta/okta-sdk-java/pull/197/files#diff-51ae4451c073ee1f09052036df605180R309

bdemers: Though thinking about it again, the message could be better 🤔

agiles231:
As a solution to rate limiting, I have implemented proktasy (https://github.com/agiles231/proktasy). Proktasy is a proxy between clients and Okta's REST API. It takes care of all rate limiting issues and handles a few edge cases that an API like this is incapable of handling. Feel free to check it out.

bdemers:
Cool idea @agiles231, I like proxy servlet approach!

How does the project protect against a single point of failure?

agiles231:
It doesn't really. If the proxy were to experience issues then all clients
would inherit those issues. However, if a transient error occurred, you
could retry just like you would when hitting Okta directly. I have thought
about adding an option to store state in a file or database in order to
share state across multiple instances, but I haven't found it worth my time
just yet (I imagine it will involve file locking and be a bit of a mess).
If multiple instances could share state, then they could be load balanced
and avoid single point of failure.

PR #205, opened 2018-07-20
Fix edge case where retry wait could be negative

bdemers:


httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -429,21 +429,32 @@ private URI getRedirectUri(HttpResponse httpResponse) {
...
 
-        // default / fallback strategy
+        // default / fallback strategy (backwards compatible implementation)
         if (delay < 0) {
             delay = Math.min(getDefaultDelayMillis(retries), timeElapsedLeft);
         }
 
+        // this shouldn't happen, but guard against a negative delay at this point
+        if (delay < 0) {
+            throw failedToRetry();
```
nbarbettini: I'm assuming this means the call fails?

bdemers: correct the `failedToRetry()` method just returns a new exception

httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -454,6 +465,10 @@ private void pauseBeforeRetry(int retries, HttpResponse httpResponse, long timeE
         }
     }
 
+    private RestException failedToRetry() {
+        return new RestException("Cannot retry request, next request will exceed retry configuration.");
```
bretterer: May be nice to give some more context here....

"Cannot retry request, next request will exceed the retry configuration of XX.  You have been retrying for XX.

Not a blocker though

httpclient/src/test/groovy/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutorTest.groovy
```diff
@@ -277,13 +278,26 @@ class HttpClientRequestExecutorTest {
 
         // maxElapsed enabled
         requestExecutor.maxRetries = 0
-        requestExecutor.maxElapsedMillis = 15000
+        requestExecutor.maxElapsedMillis = 15000L
```
bretterer: is the `L` a typo?

PR #208, opened 2018-07-24
Adds slack notifications

vijetmahabaleshwar-okta:


.travis.yml
```diff
@@ -25,4 +25,14 @@ deploy:
...
     on:
       branch: master
-      condition: $TRAVIS_EVENT_TYPE != cron
\ No newline at end of file
+      condition: $TRAVIS_EVENT_TYPE != cron
+
+notifications:
+  slack:
+    on_pull_requests: false
+    on_success: always
```
bdemers: we should probably drop `on_success` too, (to reduce the noice)

vijetmahabaleshwar-okta: @bdemers - I was just testing this PR out. Currently it's only posting to me privately on slack. Will remove it once it's working. (For some reason, I don't see the notifications on slack yet 😢)

PR #218, opened 2018-09-04
Pretty print the discriminator yaml file

bdemers:


bdemers:
NOTE: this will pass once #217 gets in

Issue #212, opened 2018-08-03
DefaultUserBuilder setProvider has no impact


PaulStat:
It seems the default implementation of the UserBuilder interface has a method for setProvider(Boolean provider) which is then used to call the client with this flag.

However as the builder has no way to set the authentication provider, when the call to client create is made the provider flag is effectively redundant as the builder hasn't set the details in the body of the user.

bdemers:
It looks like we need a `setCredentials()` object (or maybe a `setAuthenticationProvider()` that takes care of setting all three of those fields? 🤔

PaulStat:
I'd go with setCredentials, could do with adding interface for hashed password imports as well.

bdemers:
@PaulStat I went with `setCredentials()` to help support the general use cases you mentioned, password imports, etc.

But... I'd like to flatten it out, so please chime in on #226 if interested

bdemers:
@PaulStat, (I wasn't happy with that PR)
Which payloads are you trying to send, we should be able to target this use case more specifically, I'm assuming something like:

```json
"credentials": {
    "provider": {
      "type": "FEDERATION",
      "name": "FEDERATION"
    }
  }
```

and:

```json
"credentials": {
    "password" : {
      "hash": {
        "algorithm": "BCRYPT",
        "workFactor": 10,
        "salt": "rwh3vH166HCH/NT9XV5FYu",
        "value": "qaMqvAPULkbiQzkTCWo5XDcvzpk8Tna"
      }
    }
  }
```

bdemers:
(clicked the wrong button, again...)

PaulStat:
That's exactly it except with sha-256 instead of bcrypt in my case

PaulStat:
Second one imo

PR #226, opened 2018-09-24
Add `setCredentials` to UserBuilder

bdemers:
Allows for password imports and setting providers while using the UserBuilder

Fixes: #212


bdemers:
First stab at allowing setting generic credentials to be set via the UserBuilder.  

IMHO, this makes the UserBuilder a bit too complex as it all of the other setter options are _simple_.  Maybe we should include methods set these nested properties too (similar how `UserBuilder.setPassword(...)` works today

ykrysko:
Could someone please advise why wasn't this feature implemented?

PR #229, opened 2018-10-09
Adding an OkHttp based request executor (http client)

bdemers:
Retry logic was pulled out of exisitng httpclient impl and move into RetryRequestExecutor
This was done in a backwards compatible way, but ideally we only want to support a single client (one that can support a _regular_ JVM and Android)

NOTE: usage documentation has not been updated to reflect this new OkHttp client (as the existing one still works)

Either way, the splitting up of the RetryRequestExecutor makes it MUCH easier to support different implementations.


bdemers:
Update, based on #248 it looks like we have a duplicate `Timer` class, this should be resolved before merging

PR #248, opened 2019-01-10
Remove ZoneDateTime which only added in Android API 26

igormartsekha:


bdemers:
Interesting that not running the ITs only dropped coverage by ~7% (the ITs are not run against branches from external PRs)

PR #253, opened 2019-01-15
Convert swagger json to yaml

bdemers:
This is to help with view diffs moving forward

### DIFF
--- a/src/swagger/api.yaml
+++ b/src/swagger/api.yaml
@@ -13,6889 +13,4542 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
-{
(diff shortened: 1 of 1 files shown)


src/swagger/api.yaml
```diff
@@ -13,6889 +13,4542 @@
...
-        "summary": "Activate Factor",
-        "tags": [
-          "UserFactor"
-        ]
-      }
-    },
-    "/api/v1/users/{userId}/factors/{factorId}/verify": {
-      "post": {
-        "consumes": [
-
```
laura-rodriguez: isn't `>-` a typo?

PR #254, opened 2019-01-15
Add support for Policies API and ITs

### LINKED ISSUE #250: Role.type should NOT be readOnly
This prevents role.setType() from being exposed

bdemers:
Also fixes #250

### DIFF
--- a/integration-tests/src/test/groovy/com/okta/sdk/tests/it/PolicyRulesIT.groovy
+++ b/integration-tests/src/test/groovy/com/okta/sdk/tests/it/PolicyRulesIT.groovy
@@ -217,7 +217,7 @@ class PolicyRulesIT implements CrudTestSupport {
     }
 
     @Test
-    def createOktaSignOnCloudPolicyRule() {
+    void createOktaSignOnCloudPolicyRule() {
 
         def policy = client.createPolicy(client.instantiate(OktaSignOnPolicy)
             .setName("policy+" + UUID.randomUUID().toString())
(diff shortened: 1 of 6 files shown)


integration-tests/src/test/groovy/com/okta/sdk/tests/it/PolicyRulesIT.groovy
```diff
@@ -0,0 +1,298 @@
...
+                    .setSession(client.instantiate(OktaSignOnPolicyRuleSignonSessionActions)
+                        .setUsePersistentCookie(false)
+                        .setMaxSessionIdleMinutes(720)
+                        .setMaxSessionLifetimeMinutes(0))))
+            .setConditions(client.instantiate(OktaSignOnPolicyRuleConditions)
+                .setNetwork(client.instantiate(PolicyNetworkCondition)
+                    .setConnection(PolicyNetworkCondition.ConnectionEnum.ANYWHERE))
+                .setAuthContext(client.instantiate(PolicyRuleAuthContextCondition)
+                    .setAuthType(PolicyRuleAuthContextCondition.AuthTypeEnum.RADIUS))
+                .setPeople(client.instantiate(PolicyPeopleCondition)
```
laura-rodriguez: AuthType should be `ANY` for this scenario.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/PolicyRulesIT.groovy
```diff
@@ -0,0 +1,298 @@
...
+
+class PolicyRulesIT implements CrudTestSupport {
+
+    private OktaSignOnPolicy crudTestPolicy
+
+    @Override
+    def create(Client client) {
+
+        crudTestPolicy = client.createPolicy(client.instantiate(OktaSignOnPolicy)
+            .setName("policy+" + UUID.randomUUID().toString())
```
laura-rodriguez: We agreed some time ago to add an SDK identifier to the resources to trace possible issues from the Okta Org.
Is it still valid @vijetmahabaleshwar-okta ?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/PolicyRulesIT.groovy
bdemers: 🤔 i would have expected this test to fail

bdemers: Doh, the test has a return type, so this test wasn't being run.  Looks like it was the only on though.
Good catch!

integration-tests/src/test/groovy/com/okta/sdk/tests/it/PolicyRulesIT.groovy
bdemers: @laura-rodriguez @vijetmahabaleshwar-okta lets define this separately and I can find/replace these all (along with groups, apps, factors, etc)

PR #261, opened 2019-02-11
OWASP incorrectly things json-schema-core is processing

bdemers:
Adds exclusion to fix this in owasp-supporession.xml


wje600:
Looks good.

PR #272, opened 2019-03-08
Adding support for custom headers in InternalDataSource

bdemers:


impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -474,6 +496,15 @@ protected void applyDefaultRequestHeaders(Request request) {
         }
     }
 
+    private void applyUserAgent(String userAgentString, Request request) {
+
+        if (request.getHeaders().containsKey(USER_AGENT)) {
```
bretterer: What is this kind of thing going to do with reporting.   how do we know when to look at `USER_AGENT` vs `OKTA_USER_AGENT`. 

Should we just force it to always be `OKTA_USER_AGENT`?

bdemers: I guess this pattern is already used for the widget/auth-js. In reality, this is probably only going to be an issue for okta-auth-java, which we do expect developer to override the user-agent with the equivalent `User-Agent` header from a browser request

PR #274, opened 2019-03-21
Fixing edge case where the RetryRequestExecutor returns a null response when retry is disabled

bdemers:
* updated to latest httpclient


THIRD-PARTY-NOTICES
```diff
@@ -13,7 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 This project includes:
-  Apache Commons Codec under The Apache Software License, Version 2.0
+  Apache Commons Codec under Apache License, Version 2.0
```
bdemers: generated file

PR #275, opened 2019-03-28
Fixed regression in HttpClientRequestExecutor failing to retry on SocketException

bdemers:
When the RetryRequestExecutor was pulled out of HttpClientRequestExecutor in 1.4 SocketExceptions and SocketTimeoutExceptions were no longer retried
marked them retryable in the OkHttp implementation

Add configuration for HttpClient's "ValidateAfterInactivity" default of ~5~ 2 seconds (2 is http client's default)

The original implementation of HttpClientRequestExecutor predates this functionality

Updated to set the connection TTL of 5 minutes (which is Okta's DNS ttl). So what we end up with is connections will be reused for 5 minutes, and re-validated every 2 seconds of idle time
(updated after chatting with @cbarbara-okta and @bradhaverstein-okta)

### DIFF
--- a/httpclients/httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
+++ b/httpclients/httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
@@ -62,6 +62,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
+import java.util.concurrent.TimeUnit;
 
 /**
  * {@code RequestExecutor} implementation that uses the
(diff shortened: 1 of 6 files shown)


bdemers:
CC: @iancorcoran-okta
(looks like you are not on the reviewers list yet, but please informally vote until we get that figured out)

bdemers:
@laura-rodriguez The last constructor arg here: https://github.com/okta/okta-sdk-java/pull/275/files#diff-473de91e305f06d7154744e7c7399f64R263
(defaults to `false` if missing)

httpclients/httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -76,66 +77,34 @@
...
 
     private static final int DEFAULT_MAX_CONNECTIONS_TOTAL = Integer.MAX_VALUE;
     private static final String MAX_CONNECTIONS_TOTAL_PROPERTY_KEY = "com.okta.sdk.impl.http.httpclient.HttpClientRequestExecutor.connPoolControl.maxTotal";
-    private static final int MAX_CONNECTIONS_TOTAL;
+    private static final int MAX_CONNECTIONS_TOTAL = parseConfigValue(MAX_CONNECTIONS_TOTAL_PROPERTY_KEY,
+                                                                      DEFAULT_MAX_CONNECTIONS_TOTAL,
+                                                              "Bad max connection total value");
 
-    private static final int DEFAULT_CONNECTION_VALIDATION_INACTIVITY = 5000; // 5sec
+    private static final int DEFAULT_CONNECTION_VALIDATION_INACTIVITY = 2000; // 2sec
```
laura-rodriguez: why this default changed?

httpclients/httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -336,4 +305,18 @@ private HttpHeaders getHeaders(HttpResponse response) {
 
         return headers;
     }
+
+    private static int parseConfigValue(String key, int defaultValue, String warning) {
+
+        int configuredValue = defaultValue;
+        String configuredValueString = System.getProperty(key);
+        if (configuredValueString != null) {
```
laura-rodriguez: should this `if` be null or empty?

httpclients/httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
bdemers: 2s is the default of the underlying library, so I figured that would be a better starting point, I also think setting the max TTL will have bigger effect on limiting stale connections (so this now just exposes the configuration to the developer)

httpclients/httpclient/src/main/java/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
bdemers: `null`, as it isn't valid key and should stay unset

httpclients/httpclient/src/test/groovy/com/okta/sdk/impl/http/httpclient/HttpClientRequestExecutorStaticConfigTest.groovy
```diff
@@ -19,56 +19,85 @@ import org.testng.IHookCallBack
...
+    Object[][] timeToLive() {
+        return standardConfigTests(300000)
+    }
+
+    static Object[][] standardConfigTests(int defaultValue) {
+        return [
+               ["", defaultValue],
+               ["0", 0],
+               ["-1", -1],
+               ["😊", defaultValue],
```
laura-rodriguez: :)

PR #288, opened 2019-06-10
Add password hash methods to UserBuilder

bdemers:
Add import password setter for SHA-256, SHA-512, and BCRYPT

Fixes: #212

### DIFF
--- a/impl/src/main/java/com/okta/sdk/impl/resource/DefaultUserBuilder.java
+++ b/impl/src/main/java/com/okta/sdk/impl/resource/DefaultUserBuilder.java
@@ -180,6 +180,11 @@ public class DefaultUserBuilder implements UserBuilder {
 
         // user password
         if (password != null && password.length > 0) {
+
+            if (passwordHashProperties != null) {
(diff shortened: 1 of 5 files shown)


impl/src/main/java/com/okta/sdk/impl/resource/DefaultUserBuilder.java
```diff
@@ -168,28 +170,66 @@ private User build(Client client) {
...
+            PasswordCredential passwordCredential = client.instantiate(PasswordCredential.class);
+            createCredentialsIfNeeded(user, client).setPassword(passwordCredential.setValue(password));
+        }
 
-            if (password != null && password.length > 0) {
-                PasswordCredential passwordCredential = client.instantiate(PasswordCredential.class);
-                credentials.setPassword(passwordCredential.setValue(password));
-            }
+        // password import
+        if (passwordHashProperties != null) {
```
iancorcoran-okta: Thoughts on erroring if both a password and a hash are specified? Unlikely to happen, but looks like it would silently overwrite the PasswordCredential created above.

PR #324, opened 2019-10-14
Updates the jackson-databind version

vijetmahabaleshwar-okta:
Updates the jackson-databind version to 2.1.4

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -75,7 +75,7 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-        <groupId>org.powermock</groupId>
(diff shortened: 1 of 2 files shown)


vijetmahabaleshwar-okta:
> use `2.10.0` instead

Using this version is failing some tests - https://travis-ci.org/okta/okta-sdk-java/jobs/597783357

bdemers:
I'm guessing it _should_ look like: #325
(I hacked this in through the GitHub UI, so I'm not 100% sure that will pass either)
The last time we updated the jackson version it was an odd partial release that didn't update all of the individual jackson projects, going forward, we _should_ be able to just update that property

PR #339, opened 2020-02-03
WIP: Support OAuth access token ClientCredentials impl

bdemers:
Quick attempt supporting access token for authentication.
It basically _just works_ but we probably need a sperate set of tests
The current ITs set up a lot of structures and require more access than what they are actually testing.

There is no way for the SDK to obtain an access token (which is fine, that would likely be configured through some other integration)

This impl also depends on a ThreadLocal usage which may or may not be correct depending on the use case, so we would need to document (and likely provide a wrapper) to make sure any access tokens were cleaned up for the thread, maybe something like:

```java
client.withAccessToken(accessToken, (client) -> {
  // do stuff with the client here
  client.getUsers();
} );
```

### DIFF
--- a/api/src/main/java/com/okta/sdk/authc/credentials/OAuth2ClientCredentials.java
+++ b/api/src/main/java/com/okta/sdk/authc/credentials/OAuth2ClientCredentials.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2020-Present Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.sdk.authc.credentials;
+
+public class OAuth2ClientCredentials implements ClientCredentials<String> {
+
+    private static ThreadLocal<String> ACCESS_TOKEN = new ThreadLocal<>(); // TODO this might not be the best strategy, need to make sure we doc how to use and explain the importance of cleaning up after a request
+
+    @Override
+    public String getCredentials() {
+        return ACCESS_TOKEN.get();
+    }
+
+    public void setAccessToken(String accessToken) {
+        this.ACCESS_TOKEN.set(accessToken);
+    }
+
+    public void clearAccessToken() {
+        this.ACCESS_TOKEN.remove();
+    }
+}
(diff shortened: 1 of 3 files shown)


bo0tzz:
I'd really like to use this feature. Is there anything I can do to help it along?

bdemers:
@bo0tzz Great!

IMHO, The biggest gap we have on this is data ;)

There are a few different ways to get access tokens, and it depends on the type of application you are building.
The SDK currently supports using a PrivateKey exchange (which will create an OAuth2 access token for you).

Then there is this idea, which is a "bring your own access token", this would be for clients that need to make requests on behalf of a user (though in these cases caching should likely be disabled)

Lastly, there is a client credentials option ,in which you would potentially configure the SDK with an OAuth client-id, client-secret (and likely a set of OAuth scopes), and again the SDK would manage the actual token exchange for you.

The actual difference is usually one of access control, are you getting resources that only a particular user has access to? Or are you just displaying resources that all your users have access to.

bo0tzz:
I'm on the BYO path, using a stored access token to make a request against the API (okta.users.read scope). As far as I can tell the changes in this PR support what I want to do.

bdemers:
@bo0tzz you _should_ be able to lift the `ClientCredentials` implementation and include it into your code.
(you should disable caching too)

A _proper_ Spring integration would look a little different.

You might want to inject the `OAuth2AuthorizedClient`, and pull the access token from that (instead of using a thread local)

Maybe a request scoped bean (I didn't test this):
```java
@Bean
@RequestScope
class ClientCredentials oktaSdkClientCredentials(@RegisteredOAuth2AuthorizedClient("okta") OAuth2AuthorizedClient authorizedClient) {
  return () -> authorizedClient.getAccessToken().getValue();
}
```

And then configure the SDK client with that.

If you want to hack around with it, you can send a PR over to the `sdk` module of okta/okta-spring-boot: https://github.com/okta/okta-spring-boot/tree/master/sdk

bo0tzz:
I already looked into lifting this implementation, but I couldn't find a way, since the ClientBuilder will only accept RequestAuthenticators that are defined in the AuthenticationScheme enum. If there's a way around that, I'd love to hear it.

bdemers:
You _should_ be able to call `clientBuilder.setClientCredentials(...)` and ignore the `AuthenticationScheme`

**NOTE:** the latest version 2.0 replaces `AuthenticationScheme` to `AuthenticationMode`, but is basically the same code, and feel free to target the 1.x or 2.x (the logic is the same)

bo0tzz:
I was able to do this after casting the ClientBuilder so I could access the setRequestAuthenticatorFactory. I also needed to implement my own RequestAuthenticator to properly set the Authorization header. Thanks for the help!

bdemers:
@bo0tzz 

Great!

(Ideally, the API shouldn't force you to cast, but that should be something we can improve)

Thinking about caching a little more, I mentioned it should be disabled (as you are sharing the client with multiple users), we could probably improve this by updating the caches key to include a hash of the request info (specifically the auth header).
Caching may also be less of an issue in these use cases as (I'm guessing) you are making fewer read requests that need to be repeated?
I'd love to hear more about your actual use-case, either here or via email `<first.last _at_ okta.com>`

Keep us posted!

karlsequeira:
Any update on this. Would really like to use the SDK with access token

arvindkrishnakumar-okta:
Closing this, the SDK has the [capability](https://github.com/okta/okta-sdk-java/pull/841) to be used with a custom access token.

PR #342, opened 2020-02-19
Disables a failing test from CI

vijetmahabaleshwar-okta:


integration-tests/src/test/groovy/com/okta/sdk/tests/it/UsersIT.groovy
```diff
@@ -370,7 +370,7 @@ class UsersIT extends ITSupport implements CrudTestSupport {
         client.getUser(user.getId())
     }
 
-    @Test
+    @Test (groups = "bacon")
```
vijetmahabaleshwar-okta: This test will not be run on bacon as it depends on a particular monthly flag being enabled/disabled.
The dev team plan to fix this by code freeze this week. But until then, running this will fail on PDV (for orgs that don't have this flag enabled). Hence disabling this test (tests with "bacon" annotation won't run)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRulesIT.groovy
```diff
@@ -125,7 +126,7 @@ class GroupRulesIT implements CrudTestSupport {
         // 2. Create a group rule and verify rule executes
         GroupRule rule = client.instantiate(GroupRule)
                 .setType('group_rule')
-                .setName('Test group rule')
```
vijetmahabaleshwar-okta: Static name causing issues

PR #343, opened 2020-02-23
Fix auth context

BadLambdaJamma:
Fixed two child properties of LogEvent AuthenticationContext that had been set to array instead of string. (credential provider and credential type). Changed definition in api.yaml. Added a new stub file that contains log events with these items populated as observed from a running Okta tenant. Added to the existing unit test to ensure these two properties are being generated properly.


bdemers:
@BadLambdaJamma Thanks!

This looks great! And thanks for adding a test!
I'll pass this around the team (and the other SDKs).  This is a breaking (technically) change (even if this was an issue with the swagger spec).  In the short term, you should be able to call `get("credentialType", String.class)`

arvindkrishnakumar-okta:
@BadLambdaJamma With the recent major release version [2.0.0](https://github.com/okta/okta-sdk-java/releases/tag/okta-sdk-root-2.0.0), we have updated `credentialProvider` & `credentialType` to enum types.

See https://github.com/okta/okta-sdk-java/blob/master/src/swagger/api.yaml#L8963 & https://github.com/okta/okta-sdk-java/blob/master/src/swagger/api.yaml#L8965.

Feel free to reopen this issue if you have any followup questions.

PR #355, opened 2020-04-02
adds builders for Policy and its types, PolicyRule and types, GroupRule. Edited IT to use these builders.

jeyadevvishnu:
Adds builders for Policy and PasswordPolicy


bdemers:
Great stuff @jeyadevvishnu !
If you haven't already can you sign the CLA: https://github.com/okta/okta-sdk-java/blob/master/CONTRIBUTING.md#sign-the-cla

api/src/main/java/com/okta/sdk/resource/policy/PolicyBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.sdk.resource.policy;
+
+import com.okta.commons.lang.Classes;
+import com.okta.sdk.client.Client;
+
+
+public interface PolicyBuilder {
```
bdemers, line 22: This should probably be something like `PolicyBuilder<T extends PolicyBuilder>` 
and then all of the methods should return `T`.

And the corresponding above `PasswordPolicyBuilder` would be:
```java
public interface PasswordPolicyBuilder extends PolicyBuilder<PasswordPolicyBuilder> { ... }
```

In the `PolicyBuilder` you would likely need to cast to `(T)`, and to avoid all of the cast warnings you can do something like this: https://github.com/paseto-toolkit/jpaseto/blob/master/impl/src/main/java/dev/paseto/jpaseto/impl/AbstractPasetoBuilder.java#L53-L59

impl/src/main/java/com/okta/sdk/impl/resource/DefaultPasswordPolicyBuilder.java
```diff
@@ -0,0 +1,368 @@
+/*
+ * Copyright 2017 Okta
```
bdemers, line 2: nit: this should be `2020 Okta, Inc.`
(I personally take the blame for not updating the header template)
@jeyadevvishnu don't worry about his one, we can do it if you don't get a chance.

bdemers:
@jeyadevvishnu Looks like CI failed with a code check,
If you run the build with `-Pci`it will run them
(or just run the goals `pmd:check` or `findbugs:check` IIRC)

Again, great work!

api/src/main/java/com/okta/sdk/resource/policy/PolicyBuilder.java
jeyadevvishnu: Adding generic will break DefaultPasswordPolicyBuilder.class invocation from within the PasswordPolicyBuilder interface. How do we go around that in this case? Please advise.

impl/src/main/java/com/okta/sdk/impl/resource/DefaultPasswordPolicyBuilder.java
jeyadevvishnu: I will correct these with the upcoming commit. Thank you for pointing it out.

api/src/main/java/com/okta/sdk/resource/policy/PolicyBuilder.java
bdemers: Sorry, I don't understand the question.
NOTE: my suggestions were from memory, so I very well could have missed something

jeyadevvishnu: No problem at all, when we add generics as suggested up here, we below instantiation of the implementation from the `PasswordPolicyBuilder` will fail. So the question was, how do we get around it.

` static PasswordPolicyBuilder instance() {
        return Classes.newInstance("com.okta.sdk.impl.resource.DefaultPasswordPolicyBuilder");
    }`

bdemers: What is the runtime exception?

jeyadevvishnu: com.okta.commons.lang.InstantiationException: Unable to instantiate class [com.okta.sdk.impl.resource.DefaultPasswordPolicyBuilder]

impl/src/main/java/com/okta/sdk/impl/resource/DefaultPasswordPolicyBuilder.java
arvindkrishnakumar-okta: I shall update the template shortly.

api/src/main/java/com/okta/sdk/resource/policy/PolicyBuilder.java
jeyadevvishnu: Awesome! Thank you!

bdemers:
Don't worry about the coverage, the ITs don't run against remote forks.

jeyadevvishnu:
Thanks a lot for the comments and support.

jeyadevvishnu:
Also, I have signed and emailed the CLA as you mentioned.

api/src/main/java/com/okta/sdk/resource/policy/rule/SignOnPolicyRuleBuilder.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2020 Okta
```
arvindkrishnakumar-okta, line 2: Copyright 2020-Present Okta, Inc.

jeyadevvishnu: Corrected and committed. Thank you!

integration-tests/src/test/groovy/com/okta/sdk/tests/it/ApplicationsIT.groovy
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Okta
+ * Copyright 2020-Present Okta, Inc.
```
bdemers, line 2: nit: the original date of the copyright is actually the important one ;)

bdemers:
Great stuff!!

We are wrapping up a couple of other features right now, but we should be able to pull these in next!

integration-tests/src/test/groovy/com/okta/sdk/tests/it/ApplicationsIT.groovy
bdemers: also, @jeyadevvishnu, we can tweak it before merging, so don't worry too much about it 

Thanks again!

jeyadevvishnu: Thank you @bdemers, still not clear of what should be in the header, so I will leave it up to you guys. :)

jeyadevvishnu:
Thank you!

integration-tests/src/test/groovy/com/okta/sdk/tests/it/ApplicationsIT.groovy
bdemers: We are on it, I know there was a little confusion, with the header formats (which was my fault)

bdemers:
```suggestion
 * Copyright 2017-Present Okta, Inc.
```

PR #362, opened 2020-04-17
Adds encrypted private key to use in OAuth2 ITs

vijetmahabaleshwar-okta:
* Encrypts private key needed for OAuth ITs
* Will be available after decryption at `/tmp/privateKey.pem`
* Private key is specific to service app in the org we use for running travis tests (https://java-sdk.oktapreview.com)


vijetmahabaleshwar-okta:
> I'd suggest storing the key directly on travis, and not in the source tree at all. (they are encrypted/decrypted automatically)

Do you mean I should store the value of the private key into a travis variable?
Java SDK only accepts path to the private key pem file as env var. So, I'll have to write the variable to a file each time while running tests. 
That's why I wanted to directly store the file in travis.

bdemers:
Oh right, sorry, we should probably just fix that. But I'm not going to block this (strong pref against adding secrets when possible)

PR #383, opened 2020-05-27
OKTA-289655: Add Admin role API Integration Tests

arvindkrishnakumar-okta:


integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Tests for /api/v1/groups/roles
+ * @since 2.0.0
+ */
+class GroupRolesIT extends ITSupport implements CrudTestSupport {
+
+    @Override
+    def create(Client client) {
+        return GroupBuilder.instance()
```
bdemers, line 45: Is this creating a Group? or a GroupRole?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/UserRolesIT.groovy
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Tests for /api/v1/users/roles
+ * @since 2.0.0
+ */
+class UserRolesIT extends ITSupport implements CrudTestSupport {
+
+    @Override
+    def create(Client client) {
+        def email = "joe.coder+${uniqueTestName}@example.com"
+        User user = UserBuilder.instance()
```
bdemers, line 46: Same with this one, is this creating a User or a UserRole?

You don't need to implement CrudTestSupport if you want to run Resource specific CRUD tests like creating a new UserRole, (i'm guessing in the case you would create a use, and the User role from the `create` method

arvindkrishnakumar-okta: Fixed.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
arvindkrishnakumar-okta: Got rid of this as per your other comment.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
```diff
@@ -0,0 +1,213 @@
...
+
+        // 4. Assign USER_ADMIN & APP_ADMIN roles
+        Role userAdminRole = createdGroup.assignRole(userAdminRoleRequest)
+        Role appAdminRole = createdGroup.assignRole(appAdminRoleRequest)
+
+        // 5. List assigned group roles and check the assignments
+        RoleList roles = client.listGroupAssignedRoles(createdGroup.getId())
+
+        assertRolePresent(roles, userAdminRole)
+        assertRolePresent(roles, appAdminRole)
```
laura-rodriguez, line 72: I guess you have a cleanup somewhere same as apps?

laura-rodriguez: Nvm, I just saw `registerForCleanup` above :).

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
```diff
@@ -0,0 +1,213 @@
...
+        registerForCleanup(createdGroup)
+
+        validateGroup(createdGroup, groupName)
+
+        // 2. List all groups and find the group created
+        assertGroupPresent(client.listGroups(), createdGroup)
+
+        // 3. Create assign role requests
+        AssignRoleRequest userAdminRoleRequest = client.instantiate(AssignRoleRequest)
+        userAdminRoleRequest.setType(RoleType.USER_ADMIN)
```
bdemers, line 59: This is an artifact of generated code, but it _would be nice_ if this looked something like `createdGroup.assignRole(RoleType.USER_ADMIN)`.
(this would require a custom "Group" template which we may not want to maintain (but it would make the code more clear) 🧂

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
```diff
@@ -0,0 +1,213 @@
...
+
+        // 4. Assign the above role to the first group
+        Role userAdminRole = createdGroup1.assignRole(userAdminRoleRequest)
+
+        // 5. Add second group as the admin target
+        String href = "/api/v1/groups/" + createdGroup1.getId() + "/roles/" + userAdminRole.getId() +
+            "/targets/groups/" + createdGroup2.getId() + ""
+        userAdminRole.setResourceHref(href) //TODO: find the right way of doing this!
```
laura-rodriguez, line 143: Why do you need to modify the href manually?

bdemers, line 143: https://github.com/okta/okta-sdk-java/blob/master/impl/src/main/java/com/okta/sdk/impl/resource/OktaResourceHrefResolver.java

Not all of the resources return the it's href the same way.  That said most of the time, the generated code uses the correct HREF, but there are a couple special cases, for example `/api/v1/users/joe.coder@example.com` and `/api/v1/users/<some-id>` could be the same resource.  In order for caching to work correctly the same HREF needs to be used.

That said, I'm not 100% sure of what you are trying to do. something like, without looking at the code, my guess is it would look something like `createdGroup1.getRole(userAdminRole.getId()).targetGroup(createdGroup2.getId())`

laura-rodriguez: There's a method called `addGroupTargetToGroupAdministratorRoleForGroup`, is that the one you want to use here?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
```diff
@@ -0,0 +1,213 @@
...
+        String href1 = "/api/v1/groups/" + createdGroup1.getId() + "/roles/" + userAdminRole.getId() +
+            "/targets/groups/" + createdGroup2.getId() + ""
+        userAdminRole.setResourceHref(href1) //TODO: find the right way of doing this!
+
+        userAdminRole.addAdminGroup(createdGroup2.getId())
+
+        // 6. Add third group as the admin target for the first group
+        String href2 = "/api/v1/groups/" + createdGroup1.getId() + "/roles/" + userAdminRole.getId() +
+            "/targets/groups/" + createdGroup3.getId() + ""
+        userAdminRole.setResourceHref(href2) //TODO: find the right way of doing this!
```
laura-rodriguez, line 199: Same comment above, don't you want to use `AddGroupTargetToGroupAdministratorRoleForGroup` here?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
arvindkrishnakumar-okta: True, agreed.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
arvindkrishnakumar-okta: Yeah, looks like it.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
arvindkrishnakumar-okta: Seems like it, yeah.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
arvindkrishnakumar-okta: Updated.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/GroupRolesIT.groovy
```diff
@@ -0,0 +1,195 @@
...
+        // 2. List all groups and find the groups created
+        assertGroupPresent(client.listGroups(), createdGroup1)
+        assertGroupPresent(client.listGroups(), createdGroup2)
+
+        // 3. Create a USER_ADMIN role & assign it to the first group
+        Role userAdminRole = createdGroup1.assignRole(client.instantiate(AssignRoleRequest)
+            .setType(RoleType.USER_ADMIN))
+
+        // 4. Add second group as the admin target for the user admin role
+        userAdminRole.addAdminGroup(createdGroup2.getId())
```
laura-rodriguez, line 138: My understanding is that [Add a group target to a group admin role endpoint](https://developer.okta.com/docs/reference/api/roles/#add-a-group-target-to-a-group-administrator-role) needs three inputs:
```
/api/v1/groups/${groupId}/roles/${roleId}/targets/groups/${targetGroupId}
```
where do you get group1 reference?

arvindkrishnakumar-okta: Good question. 

First, I create a group (createdGroup1) and then invoke `assignRole` operation on the created group instance.

`Role userAdminRole = createdGroup1.assignRole(client.instantiate(AssignRoleRequest)
            .setType(RoleType.USER_ADMIN))`

Next, I associate the group target (createdGroup2) to the above created admin role:

`userAdminRole.addAdminGroup(createdGroup2.getId())`

Hope this clarifies.

arvindkrishnakumar-okta: > > Good question.
> > First, I create a group (createdGroup1) and then invoke `assignRole` operation on the created group instance.
> > `Role userAdminRole = createdGroup1.assignRole(client.instantiate(AssignRoleRequest) .setType(RoleType.USER_ADMIN))`
> 
> This is clear ^.
> 
> > Next, I associate the group target (createdGroup2) to the above created admin role:
> > `userAdminRole.addAdminGroup(createdGroup2.getId())`
> 
> This is what I am missing^. How do you translate these 2 params to the next href?
> 
> ```
> /api/v1/groups/${groupId}/roles/${roleId}/targets/groups/${targetGroupId}
> ```
> 
> I can imagine you are doing this:
> 
> ```
> /api/v1/groups/${groupId}/roles/${userAdminRole.GetId()}/targets/groups/${createdGroup2.getId()}
> ```
> 
> but how do you map group 1 here? what about if you have previously assigned more than one group to the same role? How do you make sure you are adding a group target to the corresponding group?
> 

Okta core returns the `groupId` in the `href` of the response for `assignRole` operation. 
I use the groupId from there.

> In .NET we have something like this:
> 
> ```
> client.Groups.AddGroupTargetToGroupAdministratorRoleForGroupAsync(createdGroup1.Id, role.Id, createdGroup2.Id);
> ```
> 
> makes sense?
> 
> > Hope this clarifies.

laura-rodriguez: After discussing this offline, it makes sense now. My only suggestion is to rename `addAdminGroup` to something like `addAdminGroupTarget`, so it is clear you are adding a group target.

arvindkrishnakumar-okta: Updated.

PR #391, opened 2020-06-03
OKTA-289657: Add Factors API Integration Tests

arvindkrishnakumar-okta:
### Resolves

[OKTA-289657](https://oktainc.atlassian.net/browse/OKTA-289657)

### DIFF
--- a/integration-tests/src/test/groovy/com/okta/sdk/tests/it/FactorsIT.groovy
+++ b/integration-tests/src/test/groovy/com/okta/sdk/tests/it/FactorsIT.groovy
@@ -21,17 +21,27 @@ import com.okta.sdk.client.Client
 import com.okta.sdk.resource.user.User
 import com.okta.sdk.resource.user.factor.ActivateFactorRequest
 import com.okta.sdk.resource.user.factor.CallUserFactor
+import com.okta.sdk.resource.user.factor.EmailUserFactor
+import com.okta.sdk.resource.user.factor.EmailUserFactorProfile
 import com.okta.sdk.resource.user.factor.FactorProvider
 import com.okta.sdk.resource.user.factor.FactorStatus
+import com.okta.sdk.resource.user.factor.FactorType
+import com.okta.sdk.resource.user.factor.HardwareUserFactor
+import com.okta.sdk.resource.user.factor.HardwareUserFactorProfile
 import com.okta.sdk.resource.user.factor.PushUserFactor
 import com.okta.sdk.resource.user.factor.SecurityQuestionUserFactor
 import com.okta.sdk.resource.user.factor.SecurityQuestionList
 import com.okta.sdk.resource.user.factor.SmsUserFactor
+import com.okta.sdk.resource.user.factor.TokenUserFactor
 import com.okta.sdk.resource.user.factor.TotpUserFactor
+import com.okta.sdk.resource.user.factor.U2fUserFactor
+import com.okta.sdk.resource.user.factor.U2fUserFactorProfile
 import com.okta.sdk.resource.user.factor.UserFactor
 import com.okta.sdk.resource.user.factor.UserFactorList
 import com.okta.sdk.resource.user.factor.VerifyFactorRequest
 import com.okta.sdk.resource.user.factor.VerifyUserFactorResponse
+import com.okta.sdk.resource.user.factor.WebUserFactor
+import com.okta.sdk.resource.user.factor.WebUserFactorProfile
 import com.okta.sdk.tests.it.util.ITSupport
 import org.jboss.aerogear.security.otp.Totp
 import org.testng.annotations.Test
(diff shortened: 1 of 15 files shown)


integration-tests/src/test/groovy/com/okta/sdk/tests/it/FactorsIT.groovy
```diff
@@ -178,6 +188,60 @@ class FactorsIT extends ITSupport {
...
+        User user = randomUser()
+        assertThat user.listFactors(), emptyIterable()
+
+        WebUserFactor webUserFactor = client.instantiate(WebUserFactor)
+            .setFactorType(FactorType.WEBAUTHN)
+            .setProvider(FactorProvider.DUO)
+            .setProfile(client.instantiate(WebUserFactorProfile))
+
+        assertThat webUserFactor.id, nullValue()
+        assertThat webUserFactor, sameInstance(user.enrollFactor(webUserFactor)) //TODO: enable this factor in oktapreview org; until then this test will fail.
```
arvindkrishnakumar-okta, line 240: @vijetmahabaleshwar-okta May I request your help with this (assuming I don't have permissions to do the same?).

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FactorsIT.groovy
```diff
@@ -178,6 +188,60 @@ class FactorsIT extends ITSupport {
...
+        User user = randomUser()
+        assertThat user.listFactors(), emptyIterable()
+
+        U2fUserFactor u2fUserFactor = client.instantiate(U2fUserFactor)
+            .setFactorType(FactorType.U2F)
+            .setProvider(FactorProvider.GOOGLE)
+            .setProfile(client.instantiate(U2fUserFactorProfile))
+
+        assertThat u2fUserFactor.id, nullValue()
+        assertThat u2fUserFactor, sameInstance(user.enrollFactor(u2fUserFactor)) //TODO: check why Okta core responds with -  HTTP 400, Okta E0000001 (Api validation failed: factorEnrollRequest - Invalid Factor.), ErrorId oaeV_agqOzrSL-dXGwvdDLpSg
```
arvindkrishnakumar-okta, line 224: @vijetmahabaleshwar-okta Looks like the factor provider `Google` needs to be enabled in Oktapreview Org?

arvindkrishnakumar-okta:
> Not my call, but `TokenUserFactor` isn't a great name for a TOTP factor (or any factor) as "token" is a pretty generic term in the security world.
> 
> That said, I don't remember what other token types that class supports. IIRC, we had SMS and other factors (which are all token based)

Here is the entire list of User Factor types defined in spec:

`
  x-openapi-v3-discriminator:
      mapping:
        call: '#/definitions/CallUserFactor'
        email: '#/definitions/EmailUserFactor'
        push: '#/definitions/PushUserFactor'
        question: '#/definitions/SecurityQuestionUserFactor'
        sms: '#/definitions/SmsUserFactor'
        token: '#/definitions/TokenUserFactor'
        'token:hardware': '#/definitions/HardwareUserFactor'
        'token:software:totp': '#/definitions/TotpUserFactor'
        u2f: '#/definitions/U2fUserFactor'
        web: '#/definitions/WebUserFactor'
      propertyName: factorType
`
So as per your recommendation, there shouldn't be a `TokenUserFactor` at all as it is too vague a name (given that we have more specific definitions like `HardwareUserFactor`, `TotpUserFactor` et al)? Or even better, we could have `TokenUserFactor` as the super class for the other token types? In any case, it would require a significant spec update.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FactorsIT.groovy
```diff
@@ -178,6 +189,43 @@ class FactorsIT extends ITSupport {
...
+    }
+
+    @Test
+    void testTokenUserFactorCreation() {
+        User user = randomUser()
+        assertThat user.listFactors(), emptyIterable()
+
+        TokenUserFactor tokenUserFactor = client.instantiate(TokenUserFactor)
+            .setFactorType(FactorType.TOKEN_SOFTWARE_TOTP)
+            .setProvider(FactorProvider.GOOGLE)
```
bdemers, line 221: There is a little overlap with the `activateTotpFactor` for this one, maybe changing the name to include GoogleTotp would make it more clear (if that is the goal of this test)

arvindkrishnakumar-okta:
> Take a look at the above comment before merging

sure, will do.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FactorsIT.groovy
arvindkrishnakumar-okta: Makes sense, yeah that's the idea of this test.

arvindkrishnakumar-okta: renamed test to `testGoogleTotpUserFactorCreation`.

arvindkrishnakumar-okta:
> > Take a look at the above comment before merging
> 
> sure, will do.

Addressed it.

PR #395, opened 2020-06-04
OKTA-289662: Add Sms Templates API Integration Tests

arvindkrishnakumar-okta:


integration-tests/src/test/groovy/com/okta/sdk/tests/it/SmsTemplateIT.groovy
```diff
@@ -0,0 +1,87 @@
...
+
+        assertThat(smsTemplate.getId(), notNullValue())
+
+        // list templates
+        assertPresent(client.listSmsTemplates(), smsTemplate)
+
+        // retrieve template
+        SmsTemplate retrievedSmsTemplate = client.getSmsTemplate(smsTemplate.getId())
+        assertThat(retrievedSmsTemplate, notNullValue())
+        assertThat("Incorrect Translations Count", retrievedSmsTemplate.getTranslations().size() == 2)
```
bdemers, line 60: nit: use `assertThat(retrievedSmsTemplate.getTranslations(), hasSize(2))`
(from memory, so I might not be 100% correct)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/SmsTemplateIT.groovy
```diff
@@ -0,0 +1,87 @@
...
+
+        // partial update template
+        retrievedSmsTemplate.getTranslations().put("es", "\${org.name}: su código de inscripción es \${code}")
+        retrievedSmsTemplate.partialUpdate()
```
bdemers, line 64: `partialUpdate()`?
Is this actually a partial update? The SDK will likely post the full object, so this _should_ be a just `update()` to match all the other resources

integration-tests/src/test/groovy/com/okta/sdk/tests/it/SmsTemplateIT.groovy
```diff
@@ -0,0 +1,87 @@
...
+        // full update template
+        SmsTemplateTranslations newSmsTemplateTranslations = client.instantiate(SmsTemplateTranslations)
+        newSmsTemplateTranslations.put("fr", "\${org.name}: votre code d'inscription est \${code}")
+
+        retrievedSmsTemplate.setName("new-" + templateName)
+            .setType(SmsTemplateType.CODE)
+            .setTemplate("\${org.name}: your verification code is \${code}")
+            .setTranslations(newSmsTemplateTranslations)
+
+        retrievedSmsTemplate.update()
```
bdemers, line 76: oh, i see `update` is here :)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/SmsTemplateIT.groovy
arvindkrishnakumar-okta: As per https://developer.okta.com/docs/reference/api/templates/#partial-sms-template-update, any update to _some_ of the properties (in our test case it is `Any translation with non-empty/null value is updated.`) is considered to be a `partial update`.

bdemers: For this test you should probably create a new instance, and just set that property,
It looks like you are reusing the previous instance (which would be the equivalent to a regular update)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/SmsTemplateIT.groovy
arvindkrishnakumar-okta: fixed

integration-tests/src/test/groovy/com/okta/sdk/tests/it/SmsTemplateIT.groovy
arvindkrishnakumar-okta: addressed this.

PR #397, opened 2020-06-05
OKTA-289661: Add Feature API Integration Tests

arvindkrishnakumar-okta:


integration-tests/src/test/groovy/com/okta/sdk/tests/it/FeaturesIT.groovy
```diff
@@ -0,0 +1,65 @@
...
+ * Tests for /api/v1/features
+ * @since 2.0.0
+ */
+class FeaturesIT extends ITSupport {
+
+    @Test
+    void featureOperationsTest() {
+        // list features
+        FeatureList featureList = client.listFeatures()
+        assertThat(featureList, iterableWithSize(greaterThanOrEqualTo(1)))
```
bdemers, line 40: I'm not familiar with the features api, but this looks like a fragile test.
Mutating the state of features could change the way the org works. IIRC some of the changes are async, and some of them can only be enabled (and never disabled).  You would need to pick a specific feature to toggle.

This test also change the global state of an org, so it would not be possible to run two sets of test against the same org at once.
My suggestion would be to turn this into a UT (and not hit a live server, your call)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FeaturesIT.groovy
```diff
@@ -0,0 +1,65 @@
...
+            // disable it
+            // "force" ensures all dependencies are also updated, if any
+            feature.updateLifecycle("false", "force")
+            assertThat(feature.getStatus(), is(EnabledStatus.DISABLED))
+        } else {
+            // enabling it programmatically may not work for all features
+            // ref: https://developer.okta.com/docs/reference/api/features/#error-response-method-not-allowed
+            try {
+                feature.updateLifecycle("true", "force")
+                assertThat(feature.getStatus(), is(EnabledStatus.ENABLED))
```
bdemers, line 59: If you go this route, pick a specific feature by ID, and toggle it.  It's likely this test would silently fail as it is

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FeaturesIT.groovy
arvindkrishnakumar-okta: That's a fair point. Will consider moving it to a UT.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FeaturesIT.groovy
arvindkrishnakumar-okta: Toggling a feature is not straightforward as it involves numerous workflow conditions as detailed at https://developer.okta.com/img/update-ssfeat-flowchart.png. I think we are better off skipping the feature lifecycle updation tests in ITs.

bdemers: +1, can you add a UT for this? (or an IT with a mocked server)

arvindkrishnakumar-okta: Ack, will do.

arvindkrishnakumar-okta: done

arvindkrishnakumar-okta: Updated.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/FeaturesIT.groovy
arvindkrishnakumar-okta: Updated with a UT instead.

PR #399, opened 2020-06-08
OKTA-289659: Add Event Hooks Integration Tests

arvindkrishnakumar-okta:


integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -0,0 +1,254 @@
...
+class EventHooksIT extends ITSupport {
+
+    @Test
+    void createEventHookTest() {
+        String name = "event-hook-java-sdk-${UUID.randomUUID().toString()}"
+
+        EventHook createdEventHook = client.instantiate(EventHook)
+            .setName(name)
+            .setEvents(client.instantiate(EventSubscriptions)
+                .setType(EventSubscriptions.TypeEnum.EVENT_TYPE)
```
bdemers, line 48: IIRC the new open API spec should be extracting all of the ENUMs into it's own type (instead of being nested)?
cc @bretterer ?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -0,0 +1,254 @@
...
+                .setVersion("1.0.0")
+                .setConfig(client.instantiate(EventHookChannelConfig)
+                    .setUri("https://www.example.com/eventHooks")
+                    .setHeaders([ client.instantiate(EventHookChannelConfigHeader)
+                                      .setKey("X-Test-Header")
+                                      .setValue("Test header value")])
+                    .setAuthScheme(client.instantiate(EventHookChannelConfigAuthScheme)
+                .setType(EventHookChannelConfigAuthSchemeType.HEADER)
+                .setKey("Authorization")
+                .setValue("Test-Api-Key"))))
```
bdemers, line 61:
```suggestion
                    .setType(EventHookChannelConfigAuthSchemeType.HEADER)
                    .setKey("Authorization")
                    .setValue("Test-Api-Key"))))
```
nit: looks like these should be nested under config

bdemers: Idea: there is a lot of room for a "builder" helper class here.  There is a lot of static data needed to create a webhook that never changes (the channel, versions, etc).  All that is needed is:

- url
- auth (header/value)
- additional headers (optional)

Maybe something like:

```java
EventHookBuilder.instance()
    .setUrl("https://www.example.com/eventHooks")
    .setAuthorizationHeader("Test-Api-Key")
    .setHeaders(Map.of("X-Test-Header",  "Test header value"))
    .buildAndCreate(client);
```

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -0,0 +1,254 @@
...
+
+    @Test
+    void getEventHookTest() {
+        String name = "event-hook-java-sdk-${UUID.randomUUID().toString()}"
+
+        EventHook createdEventHook = client.instantiate(EventHook)
+            .setName(name)
+            .setEvents(client.instantiate(EventSubscriptions)
+                .setType(EventSubscriptions.TypeEnum.EVENT_TYPE)
+                .setItems(["user.lifecycle.create", "user.lifecycle.activate"]))
```
bdemers, line 78: Are these defined somewhere? An enum probably doesn't fit for these given that they will change over time, but maybe we could define them as static args somewhere? 🤔

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -0,0 +1,254 @@
...
+                .setVersion("1.0.0")
+                .setConfig(client.instantiate(EventHookChannelConfig)
+                    .setUri("https://www.example.com/eventHooks")
+                    .setHeaders([ client.instantiate(EventHookChannelConfigHeader)
+                                      .setKey("X-Test-Header")
+                                      .setValue("Test header value")])
+                    .setAuthScheme(client.instantiate(EventHookChannelConfigAuthScheme)
+                        .setType(EventHookChannelConfigAuthSchemeType.HEADER)
+                        .setKey("Authorization")
+                        .setValue("Test-Api-Key"))))
```
bdemers, line 90: See note about builder above (IIRC async hooks are similar, except they need the subscription list).

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -0,0 +1,254 @@
...
+        registerForCleanup(createdEventHook)
+
+        assertThat(createdEventHook.getId(), notNullValue())
+
+        sleep(2000)
+
+        assertThat(client.listEventHooks(), hasSize(listSize + 1))
```
bdemers, line 217: This is prone to error if anything else changes in the org related to hooks.  Your best bet it just check if the hook has been added to the list, something like:
``` groovy
client.listEventHooks().anyMatch { createdEventHook.getId().equals(it.id) }
```

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
bretterer: Not the case yet.  This is one of the next steps that we will be doing

bdemers:
@arvindkrishnakumar-okta OH, one other thing. It doesn't look like we have a test for assigning an IT with a policy.

We have an older test over in `okta/okta-hooks-sdk-java`: https://github.com/okta/okta-hooks-sdk-java/blob/master/integration-tests/src/test/groovy/com/okta/hooks/sdk/its/RegistrationInlineHookIT.groovy#L121-L173
that creates a policy with the inline hook.

This uses an older "non-public" API, so I _hope_ there is a newer equivalent. 


We also have an OAuth based hook IT, that uses a policy rule:
https://github.com/okta/okta-hooks-sdk-java/blob/master/integration-tests/src/test/groovy/com/okta/hooks/sdk/its/OAuthTokenInlineHookIT.groovy#L358-L364

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
arvindkrishnakumar-okta: Sure, will add a static constant.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
arvindkrishnakumar-okta: Will add a builder (aka lifesaver) :)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
arvindkrishnakumar-okta: noted.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
arvindkrishnakumar-okta: Noted.

arvindkrishnakumar-okta:
> @arvindkrishnakumar-okta OH, one other thing. It doesn't look like we have a test for assigning an IT with a policy.
> 
> We have an older test over in `okta/okta-hooks-sdk-java`: https://github.com/okta/okta-hooks-sdk-java/blob/master/integration-tests/src/test/groovy/com/okta/hooks/sdk/its/RegistrationInlineHookIT.groovy#L121-L173
> that creates a policy with the inline hook.
> 
> This uses an older "non-public" API, so I _hope_ there is a newer equivalent.
> 
> We also have an OAuth based hook IT, that uses a policy rule:
> https://github.com/okta/okta-hooks-sdk-java/blob/master/integration-tests/src/test/groovy/com/okta/hooks/sdk/its/OAuthTokenInlineHookIT.groovy#L358-L364

Noted, will add that as part of InlineHooks IT.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -15,13 +15,12 @@
  */
 package com.okta.sdk.tests.it
 
-import com.okta.sdk.impl.resource.EventHookBuilder
+import com.okta.sdk.impl.resource.DefaultEventHookBuilder
```
bdemers, line 18: Looks like this is missing a tag, I'm guessing it should be `Hooks`

bdemers: OH sorry, that is the builder, so two things 
1.) There should not be an `instance()` method on `DefaultEventHookBuilder` it should be `com.okta.sdk.resource.hooks.EventHookBuilder.instance()` (and from the test you could call either `new DefaultEventHookBuilder()` or the factory method)
2.) It's missing a package declaration

bdemers: sorry, i guess it should be something like:
`com.okta.sdk.resource.event.hook.EventHookBuilder.instance()`

bdemers: The builder interfaces are always flat (or as flat as we can make them and avoid the extra nesting of the resource classes under the covers.


Something like (removing the static values, which the builder should default to):
```java
EventHook createdEventHook = EventHookBuilder.instance()
            .setName(name)
            .setUrl("https://www.example.com/eventHooks")
            .setAuthorizationHeaderValue("Test-Api-Key")
            .addHeader("X-Test-Header", "Test header value"))
            .buildAndCreate(client);
```

arvindkrishnakumar-okta: Updated.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/UserTypesIT.groovy
```diff
@@ -15,7 +15,7 @@
  */
 package com.okta.sdk.tests.it
 
-import com.okta.sdk.resource.UserType
+import com.okta.sdk.resource.user.type.UserType
```
bdemers, line 18: user.type seems a bit redundant, maybe just tag it with `user`?
(maybe there are a bunch of other user type objects 🤷)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
```diff
@@ -0,0 +1,190 @@
...
+ * Tests for {@code /api/v1/eventHooks}.
+ * @since 2.0.0
+ */
+class EventHooksIT extends ITSupport {
+
+    @Test
+    void createEventHookTest() {
+        String name = "event-hook-java-sdk-${UUID.randomUUID().toString()}"
+
+        EventHook createdEventHook = DefaultEventHookBuilder.instance()
```
bdemers, line 45: Use `new DefaultEventHookBuilder()` here or `EventHookBuilder.instance()`

integration-tests/src/test/groovy/com/okta/sdk/tests/it/UserTypesIT.groovy
arvindkrishnakumar-okta: Agreed :) 

`UserType` and `UserTypeList` are the only classes in package `com.okta.sdk.resource.user.type`.

image

Will consider moving it to a different package (say `com.okta.sdk.resource.user`) as the `user.type` seems redundant.  Will do this as part of separate PR.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/EventHooksIT.groovy
arvindkrishnakumar-okta: Done.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/UserTypesIT.groovy
arvindkrishnakumar-okta: https://oktainc.atlassian.net/browse/OKTA-305898 created to address this improvement.

arvindkrishnakumar-okta:
> One nit, and one comment (unrelated to this PR)

addressed this.

PR #409, opened 2020-06-11
OKTA-289663: Add IdP Integration Tests

arvindkrishnakumar-okta:
## Issue(s)
[OKTA-289663](https://oktainc.atlassian.net/browse/OKTA-289663)

## Description
Add Integration Tests for IdP API.

## Category

- [ ] Bugfix
- [ ] Enhancement
- [ ] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation


arvindkrishnakumar-okta:
> If updates are supported for the IdP endpoints we should add a test for that too.
> (listing them as well)

Done, ready for review.

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultIdentityProviderBuilder.java
```diff
@@ -0,0 +1,285 @@
...
+                        .setAction(ProvisioningDeprovisionedCondition.ActionEnum.NONE))
+                    .setSuspended(client.instantiate(ProvisioningSuspendedCondition.class)
+                        .setAction(ProvisioningSuspendedCondition.ActionEnum.NONE)))
+                .setGroups(client.instantiate(ProvisioningGroups.class)
+                    .setAction(ProvisioningGroups.ActionEnum.NONE)))
+            .setMaxClockSkew(maxClockSkew)
+            .setSubject(client.instantiate(PolicySubject.class)
+                .setUserNameTemplate(client.instantiate(PolicyUserNameTemplate.class)
+                    .setTemplate(subjectTemplate))
+                .setMatchType(policySubjectMatchType))));
```
bdemers, line 281: Are all of these fields required?
For example, I would guess that the jwks config could be discovered from the `issuer`.  That means these fields could be `null`, and the payload would end up with an empty `jwks` block (which may or may not cause a problem).   This goes for much of the nested structure of this object.

I'm not saying this doesn't work, I haven't used this endpoint, but if we go this route, we should include test to validate that each valid permutation of the builder works as expected (with minimal settings). And a few that expect errors, (and make sure those errors are descriptive enough),

For example, If I want to create an IdP for Google, I likely only need the to set some sort of type ("google"), scopes, client id, and client secret. (The issue is static for most social providers, and Okta already knows them)

**NOTE:** I've only created IdPs from the UI, so I'm making some [possibly incorrect] assumptions.

arvindkrishnakumar-okta: The builder `DefaultIdentityProviderBuilder` intends to cover only the generic OIDC case detailed at https://developer.okta.com/docs/reference/api/idps/#add-generic-openid-connect-identity-provider. For the sake of clarity, I would rename this class to `DefaultOIDCIdentityProviderBuilder` or something like that. 

Like you mentioned, there are several other IdP types listed at https://developer.okta.com/docs/reference/api/idps/#add-identity-provider. We could provide separate builder classes for each type in the interest of convenience and usability.

arvindkrishnakumar-okta: Pushed rename changes mentioned above.

arvindkrishnakumar-okta:
> Can you add a minimal test for a named IdP like google? I'm assuming it should only need the type, client id/secret, and scopes

sure, will do.

arvindkrishnakumar-okta:
> > Can you add a minimal test for a named IdP like google? I'm assuming it should only need the type, client id/secret, and scopes
> 
> sure, will do.

done, updated PR.

api/src/main/java/com/okta/sdk/resource/identity/provider/GoogleIdentityProviderBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+import com.okta.sdk.client.Client;
+import com.okta.sdk.resource.policy.PolicySubjectMatchType;
+
+import java.util.List;
+
+/**
+ * Builder to add Google Identity Provider.
+ * @since 2.0.0
+ */
+public interface GoogleIdentityProviderBuilder {
```
bdemers, line 28: do we need a separate interface for this?

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultGoogleIdentityProviderBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+                        .setSuspended(client.instantiate(ProvisioningSuspendedCondition.class)
+                            .setAction(ProvisioningSuspendedCondition.ActionEnum.NONE))))
+                .setAccountLink(client.instantiate(PolicyAccountLink.class)
+                    .setFilter(null)
+                    .setAction(PolicyAccountLink.ActionEnum.AUTO))
+                .setSubject(client.instantiate(PolicySubject.class)
+                    .setUserNameTemplate(client.instantiate(PolicyUserNameTemplate.class)
+                        .setTemplate(userNameTemplate))
+                    .setMatchType(policySubjectMatchType))
+                .setMaxClockSkew(maxClockSkew)));
```
bdemers, line 140: wow, is this what the object actually looks like?

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultGoogleIdentityProviderBuilder.java
```diff
@@ -0,0 +1,144 @@
...
+import com.okta.sdk.resource.identity.provider.ProvisioningSuspendedCondition;
+import com.okta.sdk.resource.policy.IdentityProviderPolicy;
+import com.okta.sdk.resource.policy.PolicyAccountLink;
+import com.okta.sdk.resource.policy.PolicySubject;
+import com.okta.sdk.resource.policy.PolicySubjectMatchType;
+import com.okta.sdk.resource.policy.PolicyUserNameTemplate;
+
+import java.util.List;
+
+public class DefaultGoogleIdentityProviderBuilder implements GoogleIdentityProviderBuilder {
```
bdemers, line 43: If we remove the `GoogleIdentityProviderBuilder` interface, and replace it with the OIDC one, this could become an implementation detail, `IdentityProviderBuilder.instance().google()`, or maybe just `IdentityProviderBuilder.google()` and `IdentityProviderBuilder.oidc()` and return the correct implementation.  I really don't know enough about the actual structure of the objects to make a strong recommendation though, it just seems that if we need a Google one, we would also need implementations for LinkedIn and Facebook (i'm hoping the request objects all looks very similar for those),

integration-tests/src/test/groovy/com/okta/sdk/tests/it/IdpIT.groovy
```diff
@@ -0,0 +1,288 @@
...
+            .setLastName("Code")
+            .setPassword("Password1".toCharArray())
+            .buildAndCreate(client)
+        registerForCleanup(createdUser)
+
+        // create google idp
+        IdentityProvider createdIdp = GoogleIdentityProviderBuilder.instance()
+            .setName("Mock Google IdP")
+            .setScopes(["openid", "profile", "email"])
+            .setProtocolType(Protocol.TypeEnum.OAUTH2)
```
bdemers, line 253: Is there another valid protocol type? for this?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/IdpIT.groovy
```diff
@@ -0,0 +1,288 @@
...
+            .setClientId("your-client-id")
+            .setClientSecret("your-client-secret")
+            .setIsProfileMaster(true)
+            .setMaxClockSkew(120000)
+            .setUserNameTemplate("idpuser.email")
+            .setPolicySubjectMatchType(PolicySubjectMatchType.USERNAME)
```
bdemers, line 259: The policy/profile master/skew bits are important, but does this work without them?

integration-tests/src/test/groovy/com/okta/sdk/tests/it/IdpIT.groovy
arvindkrishnakumar-okta: Per https://developer.okta.com/docs/reference/api/idps/#protocol-object, `Google` IdP should use `OIDC` type. I'll correct this statement here. Not sure how this IT passes with the present `OAUTH2` setting.

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultGoogleIdentityProviderBuilder.java
arvindkrishnakumar-okta: Yeah, it isn't pretty :)

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultGoogleIdentityProviderBuilder.java
arvindkrishnakumar-okta: It is a good idea to add builders for all of https://developer.okta.com/docs/reference/api/idps/#add-identity-provider.

api/src/main/java/com/okta/sdk/resource/identity/provider/GoogleIdentityProviderBuilder.java
arvindkrishnakumar-okta: Will look at a cleaner/better way.

integration-tests/src/test/groovy/com/okta/sdk/tests/it/IdpIT.groovy
arvindkrishnakumar-okta: It seems to work without them.

api/src/main/java/com/okta/sdk/resource/identity/provider/IdentityProviderBuilder.java
```diff
@@ -0,0 +1,175 @@
...
+        OIDCIdentityProviderBuilder setMaxClockSkew(Integer maxClockSkew);
+
+        OIDCIdentityProviderBuilder setSubjectTemplate(String subjectTemplate);
+
+        OIDCIdentityProviderBuilder setPolicySubjectMatchType(PolicySubjectMatchType policySubjectMatchType);
+
+        IdentityProvider buildAndCreate(Client client);
+    }
+
+    interface GoogleIdentityProviderBuilder {
```
bdemers, line 76: It doesn't look like these interfaces are needed as they contain all the fields of `OIDCIdentityProviderBuilder` ?

Maybe just do something like:
```java
static OIDCIdentityProviderBuilder google() {
        return Classes.newInstance("com.okta.sdk.impl.resource.identity.provider.DefaultGoogleIdentityProviderBuilder");
    }

    static OIDCIdentityProviderBuilder facebook() {
        return Classes.newInstance("com.okta.sdk.impl.resource.identity.provider.DefaultFacebookIdentityProviderBuilder");
    }

    static OIDCIdentityProviderBuilder microsoft() {
        return Classes.newInstance("com.okta.sdk.impl.resource.identity.provider.DefaultMicrosoftIdentityProviderBuilder");
    }

    static OIDCIdentityProviderBuilder linkedin() {
        return Classes.newInstance("com.okta.sdk.impl.resource.identity.provider.DefaultLinkedInIdentityProviderBuilder");
    }
```

This would keep the actual implementation private (and not add an extra interface)

arvindkrishnakumar-okta: Actually, `GoogleIdentityProviderBuilder` does not have to contain every method defined in `OIDCIdentityProviderBuilder`. Currently there are few methods that are shared between these interfaces. I wanted to keep these interfaces separate as it otherwise becomes an overhead to implement unnecessary methods from `OIDCIdentityProviderBuilder`. This does not fit the inheritance design principle and therefore this duplication. From caller's perspective, a `Google` specific instance could be built easily with:
  `IdentityProvider createdIdp = IdentityProviderBuilder.google()
            .setName("Mock Google IdP")
            .setScopes(["openid", "profile", "email"])
            .setClientId("your-client-id")
            .setClientSecret("your-client-secret")
            .setIsProfileMaster(true)
            .setMaxClockSkew(120000)
            .setUserNameTemplate("idpuser.email")
            .setPolicySubjectMatchType(PolicySubjectMatchType.USERNAME)
            .buildAndCreate(client)`

`GoogleIdentityProviderBuilder` interface and its implementation detail is actually hidden from the caller. The caller still has to rely on `IdentityProviderBuilder.google()` to get a Google specific builder.

bdemers: Ahh, i see.  You could end up with a base `IdentityProviderBuilder` and extend them for OIDC, Google, etc

arvindkrishnakumar-okta: done.

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.okta.commons.lang.Classes;
+
+public interface MicrosoftIdentityProviderBuilder extends IdentityProviderBuilder<MicrosoftIdentityProviderBuilder> {
+
+    static MicrosoftIdentityProviderBuilder instance() {
+        return Classes.newInstance("com.okta.sdk.impl.resource.identity.provider.DefaultMicrosoftIdentityProviderBuilder");
+    }
+
+    MicrosoftIdentityProviderBuilder setIsProfileMaster(Boolean isProfileMaster);
```
bdemers, line 26: This field should be available for all types

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
```diff
@@ -0,0 +1,14 @@
...
+
+    MicrosoftIdentityProviderBuilder setUserNameTemplate(String userNameTemplate);
```
bdemers, line 28: I think this one is too

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultFacebookIdentityProviderBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+                        .setSuspended(client.instantiate(ProvisioningSuspendedCondition.class)
+                            .setAction(ProvisioningSuspendedCondition.ActionEnum.NONE))))
+                .setAccountLink(client.instantiate(PolicyAccountLink.class)
+                    .setFilter(null)
+                    .setAction(PolicyAccountLink.ActionEnum.AUTO))
+                .setSubject(client.instantiate(PolicySubject.class)
+                    .setUserNameTemplate(client.instantiate(PolicyUserNameTemplate.class)
+                        .setTemplate(userNameTemplate))
+                    .setMatchType(policySubjectMatchType))
+                .setMaxClockSkew(maxClockSkew)));
```
bdemers, line 88: There are a bunch of fields available in the IdP creation UI that are not available in this builder (or are hard coded here)

arvindkrishnakumar-okta: let me study the UI and compare each of these builders for missing fields.

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
arvindkrishnakumar-okta: noted, will work on refactoring this.

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
arvindkrishnakumar-okta: Noted.

impl/src/main/java/com/okta/sdk/impl/resource/identity/provider/DefaultFacebookIdentityProviderBuilder.java
arvindkrishnakumar-okta: updates done.

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
arvindkrishnakumar-okta: updated

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
arvindkrishnakumar-okta: updated

api/src/main/java/com/okta/sdk/resource/identity/provider/MicrosoftIdentityProviderBuilder.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.sdk.resource.identity.provider;
+
+import com.okta.commons.lang.Classes;
+
+public interface MicrosoftIdentityProviderBuilder extends IdentityProviderBuilder<MicrosoftIdentityProviderBuilder> {
```
bdemers, line 20: Are these interfaces needed?
It doesn't look like there are anything specific here? Maybe something like:

```java
public class IdentityProviderBuilders {


    public static OIDCIdentityProviderBuilderoidc() { ... }

    public static IdentityProviderBuilder facebook() { ... }
  
    public static IdentityProviderBuilder google() { ... }
...

}
```

arvindkrishnakumar-okta: Updated.

Issue #431, opened 2020-07-08
Question: How to specify UserType while creating new user?


venuduggireddy:
How to specify UserType while creating new user? I do not see an option to assign UserType in UserBuilder similar to addGroup while creating user? Is there any option available within the SDK?

arvindkrishnakumar-okta:
@venuduggireddy Thanks for raising this question!

The `UserBuilder` presently does not have a setter for `UserType`. As a quick solution, you could call `setUserType` method on the created `User` object. 

I will add a PR to include the missing property to the builder.

PR #432, opened 2020-07-08
Add setter for UserType property in UserBuilder

arvindkrishnakumar-okta:
## Issue(s)
https://github.com/okta/okta-sdk-java/issues/431

## Description
Add setter for `UserType` property in `UserBuilder`.

## Category

- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation

### DIFF
--- a/api/src/main/java/com/okta/sdk/resource/user/UserBuilder.java
+++ b/api/src/main/java/com/okta/sdk/resource/user/UserBuilder.java
@@ -51,6 +51,8 @@ public interface UserBuilder {
 
     UserBuilder setProvider(Boolean provider);
 
+    UserBuilder setType(UserType userType);
+
     UserBuilder setProfileProperties(Map<String, Object> profileProperties);
 
     UserBuilder putAllProfileProperties(Map<String, Object> profileProperties);
--- a/api/src/main/java/com/okta/sdk/resource/user/UserBuilder.java
+++ b/api/src/main/java/com/okta/sdk/resource/user/UserBuilder.java
@@ -53,6 +53,8 @@ public interface UserBuilder {
 
     UserBuilder setType(UserType userType);
 
+    UserBuilder setType(String userTypeId);
+
     UserBuilder setProfileProperties(Map<String, Object> profileProperties);
 
     UserBuilder putAllProfileProperties(Map<String, Object> profileProperties);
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/groovy/com/okta/sdk/tests/it/UsersIT.groovy
```diff
@@ -176,6 +178,47 @@ class UsersIT extends ITSupport implements CrudTestSupport {
...
+
+        // 2. Create a user with the User Type
+        User user = UserBuilder.instance()
+            .setEmail(email)
+            .setFirstName(firstName)
+            .setLastName(lastName)
+            .setPassword(password.toCharArray())
+            .setActive(true)
+        // See https://developer.okta.com/docs/reference/api/user-types/#specify-the-user-type-of-a-new-user
+            .setType(client.instantiate(UserType).setId(createdUserType.getId()))
```
bdemers, line 209: nit, you could probably just pass in the `createdUserType` from above instead of creating one

api/src/main/java/com/okta/sdk/resource/user/UserBuilder.java
```diff
@@ -51,6 +51,8 @@ static UserBuilder instance() {
 
     UserBuilder setProvider(Boolean provider);
 
+    UserBuilder setType(UserType userType);
```
bdemers, line 54: I'm not 100% sure on how this would be used, but you may want to also add a `UserBuilder setType(String userTypeId)` I'm [only] guessing, but the caller of this may not have an instance of `UserType`

(just an idea, 🧂)

integration-tests/src/test/groovy/com/okta/sdk/tests/it/UsersIT.groovy
arvindkrishnakumar-okta: Yes, I had thought the same until I realized that we need to supply only the type `id` in the request body (in the context of creating a new user). I had added a comment in the previous line with a link to dev doc that notes this requirement.

api/src/main/java/com/okta/sdk/resource/user/UserBuilder.java
arvindkrishnakumar-okta: Added an overloaded `setType(String userTypeId)` method in `UserBuilder`.

PR #438, opened 2020-07-22
Added UserFactorProfile common parent interface for all individual UserFactor profile types

arvindkrishnakumar-okta:
## Description

Prior to the major release 2.0.0, we had FactorProfile interface as common parent for all individual profile types (call, email, sms, totp etc).

With the 2.0.0 update, we removed the FactorProfile interface.

With this PR, I am adding a functionally equivalent interface named UserFactorProfile (inline with Factor -> UserFactor name change in major rev) as the common parent for individual UserFactor profile types.

Also submitted this change to **Open API** spec project at https://github.com/okta/openapi/pull/255.

## Category

- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>com.okta.sdk</groupId>
         <artifactId>okta-sdk-root</artifactId>
-        <version>2.0.1-SNAPSHOT</version>
+        <version>2.1.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>okta-sdk-api</artifactId>
--- a/src/swagger/api.yaml
+++ b/src/swagger/api.yaml
@@ -7656,6 +7656,7 @@ definitions:
         type: string
       phoneNumber:
         type: string
+    x-okta-parent: '#/definitions/UserFactorProfile'
     x-okta-tags:
       - UserFactor
   ChangePasswordRequest:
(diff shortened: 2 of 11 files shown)


api/pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <groupId>com.okta.sdk</groupId>
         <artifactId>okta-sdk-root</artifactId>
-        <version>2.0.1-SNAPSHOT</version>
+        <version>2.1.0-SNAPSHOT</version>
```
arvindkrishnakumar-okta, line 24: Considering this a `minor` version upgrade since we are adding a backward compatible change (compare to v2.0.0) by introducing a new parent interface `UserFactorProfile`.

src/swagger/api.yaml
```diff
@@ -11818,6 +11827,10 @@ definitions:
         web: '#/definitions/WebUserFactor'
         webauthn: '#/definitions/WebAuthnUserFactor'
       propertyName: factorType
+  UserFactorProfile:
+    properties: {}
```
bdemers, line 11831: Are there any common properties that can be moved here like `credentialId` ?
(note: i'm not sure if that is common across all of the factor profiles)

arvindkrishnakumar-okta: Actually, there is no commonality between the individual user factor profiles. Like we discussed on Slack, I will remove this from the spec and have the code copy pasted into the project.

PR #439, opened 2020-07-23
Add all Headers form HttpHeaderHolder

bdemers:
This is still an "impl" specific feature, but it gets us closer to exposing this API somehow


bdemers:
I'm not sure why that diff report is off 🤔

PR #442, opened 2020-08-03
IDEA: Lookup config file based on XdgConfig

bdemers:
see: okta/okta-commons-java#38

NOTE: tests will fail as above PR is not merged


sergiishamrai-okta (commented):
line 56:
import java.io.File;

Could be removed

bdemers:
Thanks @senthilkumaran-okta the linting (PMD in this case will catch that).

I'm mostly interested in floating the idea of using the Xdg spec to resolve configuration files.  This is something that needs to be considered across all of our SDKs (not just the Java project)

sergiishamrai-okta (approved):
Looks good

bdemers:
@sergiishamrai-okta This should be reverted until this is discussed with the other SDK owners.
see #470

PR #488, opened 2020-11-20
Allow PEM-content of private key to be passed directly

### LINKED ISSUE #486: ClientBuilder::setPrivateKey should have a variant where I can set the pem-content directly
Right now, it allows me to set a String which contains a hard file-system-path. This is not appropriate in many cloud-native environments where configuration-properties like this are loaded from a service, such as AWS Secrets Manager.

Also, it is inconvenient for the pem file to _have to_ be in a physical file. In containerized cloud-native environments access to a filesystem is not as normal as it was in the old days.

It will also be much more flexible this way. Maybe passing an `InputStream` is even better.

So please allow the pem content to be set something like this:

```java

String pemContent = readPem();// so this I take care of myself
var client = Clients.builder()
                .setPrivatePem(pemContent) // pass the whole pem, avoid reading the FS
                .setOrgUrl("https://" + oktaDomain)  // e.g. https://dev-123456.okta.com
                .setAuthorizationMode(AuthorizationMode.PRIVATE_KEY)
                .setClientId("my-id")
                .setScopes(Set.of("okta.users.read"))
                .setCacheManager(Caches.newDisabledCacheManager())
                .build();
```

hanswesterbeek:
Avoids reading from filesystem which is problematic in cloud-native environments.

## Issue(s)
Fixes #486 
## Category

- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation

### DIFF
--- a/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
+++ b/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
@@ -208,6 +208,7 @@ public interface ClientBuilder {
     String DEFAULT_CLIENT_ID_PROPERTY_NAME = "okta.client.clientId";
     String DEFAULT_CLIENT_SCOPES_PROPERTY_NAME = "okta.client.scopes";
     String DEFAULT_CLIENT_PRIVATE_KEY_PROPERTY_NAME = "okta.client.privateKey";
+    String DEFAULT_CLIENT_PRIVATE_KEY_CONTENT_PROPERTY_NAME = "okta.client.privateKeyContent";
     String DEFAULT_CLIENT_REQUEST_TIMEOUT_PROPERTY_NAME = "okta.client.requestTimeout";
     String DEFAULT_CLIENT_RETRY_MAX_ATTEMPTS_PROPERTY_NAME = "okta.client.rateLimit.maxRetries";
     String DEFAULT_CLIENT_TESTING_DISABLE_HTTPS_CHECK_PROPERTY_NAME = "okta.testing.disableHttpsCheck";
--- a/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
+++ b/api/src/main/java/com/okta/sdk/client/ClientBuilder.java
@@ -333,6 +333,19 @@ public interface ClientBuilder {
     ClientBuilder setPrivateKey(String privateKey);
 
     /**
+     * Allows passing of PEM payload for the private key directly, avoiding the
+     * need to access the filesystem directly.
+     *
+     * @see #setPrivateKey(String)
+     * @param privateKeyContent the full PEM-payload
+     * @return the ClientBuilder instance for method chaining.
+     *
+     * @throws IllegalArgumentException when {@link #setPrivateKey(String)} was called before. They are mutually exclusive
+     * @since 2.0.1
+     */
+    ClientBuilder setPrivateKeyContent(String privateKeyContent);
+
+    /**
      * Allows specifying the client ID instead of relying on the default location + override/fallback behavior defined
      * in the {@link ClientBuilder documentation above}.
      *
--- a/impl/src/main/java/com/okta/sdk/impl/client/DefaultClientBuilder.java
+++ b/impl/src/main/java/com/okta/sdk/impl/client/DefaultClientBuilder.java
@@ -353,10 +353,18 @@ public class DefaultClientBuilder implements ClientBuilder {
         Assert.notNull(clientConfiguration.getClientId(), "clientId cannot be null");
         Assert.isTrue(clientConfiguration.getScopes() != null && !clientConfiguration.getScopes().isEmpty(),
             "At least one scope is required");
-        Assert.notNull(clientConfiguration.getPrivateKey(), "privateKey cannot be null");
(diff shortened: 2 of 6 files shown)


hanswesterbeek:
This is the simplest way of adding the feature I asked. We could go for something more sophisticated (e.g. a `PrivateKeyProvider` interface with two impls, but I'd leave that kind of decision up to you.

`ClientBuilder` will throw an `IllegalArgumentException` if a user tries to set both `privateKey` and `privateKeyContent`

hanswesterbeek:
Not sure why codecov is flagging this PR. It seems to see a much larger diff than I provided.

The two files I actually changed are showing up green though. Please let me know if something is wrong.

bdemers (changes requested):
Great stuff! (don't worry about all the red in the code coverage, the ITs are NOT run for external pull requests as they do NOT have access to environment variables, we will make sure they are run before merging)


Our other SDKs allow for the file or PEM to be passed via the same config system variable/environment variable in the format of:

```
okta.client.privateKey: <some-file>/privateKey.pem | -----BEGIN PRIVATE KEY----- ...
```
(This would translate automatically to an env var of `OKTA_CLIENT_PRIVATEKEY` as well)

To see where the prop gets set you can take a look at: https://github.com/okta/okta-sdk-java/blob/master/impl/src/main/java/com/okta/sdk/impl/client/DefaultClientBuilder.java#L228

We could probably check if the prop starts with `----- BEGIN`
and then assume we have a PEM string, and then call your new `setPrivateKeyContent()` method.

We could also add type-specific methods, `setPrivateKey(Path)`,  `setPrivateKey(java.security.PrivateKey`), though the current `setPrivateKey(String)` would need to be backward compatible with files and the content (until the next major release) 🤔  (but don't worry about this bit, I'm mostly just thinking out loud)

hanswesterbeek:
I generally don't use such global config mechanisms provided by any library.

Also, letting a property take two different guises like you suggest is often a bit magical to the end user.

I'd force the user to be explicit in their intention. That also allows you to refactor more easily when required.

Anyway, I'd love to see this PR make it to 2.0.1 sometime soon as it will make my life a LOT easier when deploying my app in AWS Fargate.

hanswesterbeek:
Oh btw, if you want me to implement support for an additional prop/env var I'll be happy to that if it speeds things up.

That said, there is some talk about env vars not being a safe place for secrets. Make of that what you will :)

bdemers:
@hanswesterbeek that would be awesome if you could do that!

I have mixed feelings about the env vars as secrets too, but that is just one option. There is a hierarchy of property sources (similar to what Spring Boot does), config files, system properties, env vars, etc.

Though as you mentioned it is important for folks to understand what else has access to your secrets.  For example, setting a system property on the command line `-Dfoo=bar` could show up in a `ps`, but I've also seen shops load sys props other ways to avoid this.

arvindkrishnakumar-okta:
@hanswesterbeek Thanks for your contribution! Much appreciated! 

Can you take a quick moment to sign the CLA at https://github.com/okta/okta-sdk-java/blob/master/CONTRIBUTING.md#sign-the-cla?

hanswesterbeek:
@arvindkrishnakumar-okta CLA signed and emailed.

@bdemers Yep. In our case, we let `spring-cloud-config-aws` fetch secrets and props from AWS Secrets Manager and AWS Paramstore at boot-time. No env vars, props or external config files required :)

bdemers:
Funny you should mention that. After we get this in, I was going to test it out with our Spring Boot lib! 😄

hanswesterbeek:
@bdemers I have given it some more thought and I really [don't think it's a good idea](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/) to allow the pemContent to be set in the env-vars.

bdemers:
I understand your concern about using env vars.  However, to try to ease your mind we use similar logic to how [Spring Boot loads properties](https://docs.spring.io/spring-boot/docs/1.5.6.RELEASE/reference/html/boot-features-external-config.html)

The order (our resolution order when no using Spring Boot) is:
* Programmatic configuration
* System Properties
* Environment Variables
* ~/.okta/okta.yaml
* Classpath:
  * /okta.yaml
  * /okta.properties
  * /com/okta/sdk/config/okta.yaml
  * /com/okta/sdk/config/okta.properties

hanswesterbeek:
@bdemers Ok I added the commit to load the new prop just like all the others, along with a gentle word of warning in the javadoc.

I know Spring Boot allows this as well. Not sure how happy the maintainers are with that choice at this point. Also, that is a framework while this is a library. That said, I know the AWS SDK (also a lib) does the same thing as well. What a world :)

Can you merge this and release 2.0.1?

arvindkrishnakumar-okta:
@hanswesterbeek Thanks again! I'll review it, merge and release it in the upcoming release.

arvindkrishnakumar-okta:
@hanswesterbeek Thanks for you PR again! 

This has been fixed in latest release `3.0.1`.

hanswesterbeek:
Nice work team :)

Issue #516, opened 2021-01-18
Client is not able to interact with XML response based APIs (eg SAML meta data API)


ankitrajsingh:
https://github.com/okta/okta-sdk-java/blob/8ef2ac4aaafb189fe37196f4bd2d3c44179e22f1/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java#L480

My requirement is to fetch app's SAML metadata ([API Ref](https://developer.okta.com/docs/reference/api/apps/#preview-saml-metadata-for-application)), while hitting the API through client along with correct Accept header ("application/xml"), the client forcibly overrides it to "application/json" and the request result in BAD REQUEST error code from the API mentioning that the response format can't be "application/json", it should be "application/xml" instead.

Please let me know how can the client be enabled/configured to interact with XML based okta apis.

sergiishamrai-okta:
Thanks for the request, @ankitrajsingh 

Can you elaborate on what specifically you are trying to do?

ankitrajsingh:
I am trying to link orgs via Org2Org app, for that I need to transfer SAML credentials of Org2Org App in Spoke(say Okta OrgA) to the corresponding IDP of the Hub (say Okta OrgB).

The said API provides the SAML credentials demanded by the IDP i.e entityId, X509Cert and SSO links.

sergiishamrai-okta:
Thanks for the report @ankitrajsingh.
This is definitely a bug and we're going to put it on our roadmap.  Follow this issue for updates.

arvindkrishnakumar-okta:
Internal ref: OKTA-361894

ankitrajsingh:
Hi @sergiishamrai-okta , as a temporary work around to this I am looking to define our own client for this specific api. Anyway we can carry token handover from the Okta client to our own custom client ? Any method/mechanism which Okta client exposes for the same would be great.

arvindkrishnakumar-okta:
@ankitrajsingh Not sure I understand what you mean by "token handover". Can you elaborate on what you need?

ankitrajsingh:
@arvindkrishnakumar-okta , I am talking about Oauth tokens which gets generated via client id and private key. Okta client internally takes care of generating the auth token with this information.

But as the Okta Client can't hit the SAML api due header issue, let's say i define my own HttpClient(eg. RestTemplate).
Now this client needs the auth token passed in headers while hitting SAML api.

[...]

PR #540, opened 2021-02-12
Client is not able to interact with XML response

VitaliiTytarenko-okta:
## Issue(s)
#516 

OKTA-361894

## Description
Add possibility to retrieve an HTTP response containing a non-JSON body.

## Category
- [x] Bugfix
- [x] Enhancement
- [ ] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>com.okta.sdk</groupId>
         <artifactId>okta-sdk-root</artifactId>
-        <version>4.0.0-SNAPSHOT</version>
+        <version>3.1.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>okta-sdk-api</artifactId>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -21,7 +21,7 @@
     <parent>
         <groupId>com.okta.sdk</groupId>
         <artifactId>okta-sdk-root</artifactId>
-        <version>3.0.4-SNAPSHOT</version>
+        <version>4.0.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>okta-sdk-api</artifactId>
--- a/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
+++ b/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
@@ -16,6 +16,9 @@
  */
 package com.okta.sdk.impl.ds;
 
+import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.io.CharStreams;
 import com.okta.commons.http.DefaultRequest;
 import com.okta.commons.http.HttpHeaders;
 import com.okta.commons.http.HttpMethod;
@@ -59,7 +62,9 @@ import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -470,14 +475,24 @@ public class DefaultDataStore implements InternalDataStore {
         Map<String, Object> out = null;
 
         if (response.hasBody()) {
-            out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            if ("json".equals(response.getHeaders().getContentType().getSubtype())) {
+                out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            } else {
+                try (InputStreamReader inputStreamReader = new InputStreamReader(response.getBody(), Charsets.UTF_8)) {
+                    out = ImmutableMap.of(MediaType.TEXT_PLAIN_VALUE, CharStreams.toString(inputStreamReader));
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            }
         }
 
         return out;
     }
 
     protected void applyDefaultRequestHeaders(Request request) {
-        request.getHeaders().setAccept(java.util.Collections.singletonList(MediaType.APPLICATION_JSON));
+        if (!request.getHeaders().containsKey("Accept")) {
+            request.getHeaders().setAccept(java.util.Collections.singletonList(MediaType.APPLICATION_JSON));
+        }
 
         // Get runtime headers from http client
         Map<String, List<String>> headerMap = HttpHeadersHolder.get();
--- a/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
+++ b/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
@@ -65,6 +65,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static com.okta.commons.http.HttpHeaders.OKTA_USER_AGENT;
@@ -167,6 +168,7 @@ public class DefaultDataStore implements InternalDataStore {
         return getResource(href, clazz, null);
     }
 
+    @Override
     public <T extends Resource> T getResource(String href, Class<T> clazz, Map<String, Object> queryParameters) {
         return getResource(href, clazz, queryParameters, null);
     }
@@ -179,6 +181,19 @@ public class DefaultDataStore implements InternalDataStore {
         resource.setResourceHref(href);
 
         return resource;
+    }
+
+    @Override
+    public InputStream getRawResponse(String href, Map<String, Object> queryParameters, Map<String, List<String>> headerParameters) {
+
+        Assert.hasText(href, "href argument cannot be null or empty.");
+
+        CanonicalUri uri = canonicalize(href, queryParameters);
+        Request getRequest = new DefaultRequest(HttpMethod.GET, uri.getAbsolutePath(), uri.getQuery(), headers(headerParameters));
+        Response getResponse = execute(getRequest);
+
+        return getRawBody(getResponse)
+            .orElseThrow(() -> new IllegalStateException("Unable to obtain resource data from the API server."));
     }
 
     @SuppressWarnings("unchecked")
@@ -345,7 +360,7 @@ public class DefaultDataStore implements InternalDataStore {
                 props,
                 requestHeaders);
 
-        ResourceDataResult result = chain.filter(request);
+        DefaultResourceDataResult result = (DefaultResourceDataResult) chain.filter(request);
 
         Map<String,Object> data = result.getData();
 
@@ -467,13 +482,21 @@ public class DefaultDataStore implements InternalDataStore {
 
         Assert.notNull(response, "response argument cannot be null.");
 
-        if (response.hasBody()) {
-            return MediaType.APPLICATION_JSON.equals(response.getHeaders().getContentType())
-                ? mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap())
-                : java.util.Collections.singletonMap(MediaType.TEXT_PLAIN_VALUE, response.getBody());
+        if (response.hasBody() && MediaType.APPLICATION_JSON.equals(response.getHeaders().getContentType())) {
+            return mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
         }
 
         return java.util.Collections.emptyMap();
+    }
+
+    private Optional<InputStream> getRawBody(Response response) {
+        Assert.notNull(response, "response argument cannot be null.");
+
+        if (response.hasBody()) {
+            return Optional.ofNullable(response.getBody());
+        }
+
+        return Optional.empty();
     }
 
     protected void applyDefaultRequestHeaders(Request request) {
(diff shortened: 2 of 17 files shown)


impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -470,14 +475,24 @@ private Response execute(Request request) throws ResourceException {
         Map<String, Object> out = null;
 
         if (response.hasBody()) {
-            out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            if ("json".equals(response.getHeaders().getContentType().getSubtype())) {
```
bdemers, line 478: This probably isn't generic enough, i.e. this could be `application/ion+json`

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -16,6 +16,9 @@
  */
 package com.okta.sdk.impl.ds;
 
+import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.io.CharStreams;
```
bdemers, line 21: IIRC, Guava is an optional dependency and should not be used here (I could be wrong)

I like Guava, but IMHO, it shouldn't be used in libs as it has historically had many breaking changes

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -470,14 +475,24 @@ private Response execute(Request request) throws ResourceException {
...
 
         if (response.hasBody()) {
-            out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            if ("json".equals(response.getHeaders().getContentType().getSubtype())) {
+                out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            } else {
+                try (InputStreamReader inputStreamReader = new InputStreamReader(response.getBody(), Charsets.UTF_8)) {
+                    out = ImmutableMap.of(MediaType.TEXT_PLAIN_VALUE, CharStreams.toString(inputStreamReader));
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
```
bdemers, line 484: Throw a typed exception

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -470,14 +475,24 @@ private Response execute(Request request) throws ResourceException {
         Map<String, Object> out = null;
 
         if (response.hasBody()) {
-            out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            if ("json".equals(response.getHeaders().getContentType().getSubtype())) {
+                out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            } else {
+                try (InputStreamReader inputStreamReader = new InputStreamReader(response.getBody(), Charsets.UTF_8)) {
+                    out = ImmutableMap.of(MediaType.TEXT_PLAIN_VALUE, CharStreams.toString(inputStreamReader));
```
bdemers, line 482: IMHO, I feel this overloads this functionality in a confusing way.  If we need to expose raw HTTP related functionality, then new methods should be added for that purpose, i.e. `client.httpRaw().post("/url/")`

_Ideally_, we want to return an input stream to avoid large strings in memory (when possible)

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
VitaliiTytarenko-okta: Of course

bdemers: https://trac.tools.ietf.org/html/draft-ietf-appsawg-media-type-suffix-regs-02#section-3.1

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
VitaliiTytarenko-okta: I don't quite understand why we need new methods. The aim of this PR - is to avoid exceptions of converting different formats to Map in the process of `getResource` method https://github.com/okta/okta-sdk-java/blob/90436758263d2c597319838af77cb848b2ae6f11/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java#L175
I've updated my code to return `Collections.singletonMap(MediaType.TEXT_PLAIN_VALUE, response.getBody())`, if the content type other than json.

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
bdemers: (i bring up ion, because Okta is using it for some early access endpoints)

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
bdemers: The SDKs original purpose was to hide all of the HTTP logic, so the user only deals with resources (and all of the complexities that come with that, like JSON objects changing over time).

Returning a string in a map isn't a resource, IMHO, sticking the response body in a map isn't intuitive.

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -467,17 +482,27 @@ private Response execute(Request request) throws ResourceException {
 
         Assert.notNull(response, "response argument cannot be null.");
 
-        Map<String, Object> out = null;
+        if (response.hasBody() && MediaType.APPLICATION_JSON.equals(response.getHeaders().getContentType())) {
```
arvindkrishnakumar-okta, line 485: We could possibly get `application/ion+json` in future and this would not work.

impl/src/test/groovy/com/okta/sdk/impl/ds/DefaultDataStoreTest.groovy
```diff
@@ -69,11 +69,13 @@ class DefaultDataStoreTest {
         def response = mock(Response)
         def responseBody = '{"name": "jcoder"}'
         def defaultDataStore = new DefaultDataStore(requestExecutor, "https://api.okta.com/v1", apiKeyResolver)
+        def headers = new HttpHeaders()
```
arvindkrishnakumar-okta, line 72: Update the test with `application/ion+json` once it is added per my above comment.

arvindkrishnakumar-okta:
Caller must be able to invoke the endpoint like:

client.http().addHeaderParameter("Accept", "application/xml").get("/api/v1/apps/0oa7mivnlMxc1lJ4A0w6/sso/saml/metadata", Resource.class)

We should support this as well.

VitaliiTytarenko-okta:
@arvindkrishnakumar-okta 
Now it works like a charm 
```
InputStream response = client.http()
                .addHeaderParameter("Accept", "application/xml")
                .getRaw("/api/v1/apps/${clientId}/sso/saml/metadata");
```

integration-tests/src/test/groovy/com/okta/sdk/tests/it/ApplicationsIT.groovy
```diff
@@ -706,4 +716,65 @@ class ApplicationsIT extends ITSupport {
         // verify
         assertNotPresent(app.listScopeConsentGrants(), app.getScopeConsentGrant(oAuth2ScopeConsentGrant.getId()))
     }
+
+    @Test
+    void testExecuteAcceptJson() {
```
arvindkrishnakumar-okta, line 721:
```suggestion
    void testExecuteAcceptIonPlusJson() {
```

integration-tests/src/test/groovy/com/okta/sdk/tests/it/ApplicationsIT.groovy
```diff
@@ -706,4 +716,65 @@ class ApplicationsIT extends ITSupport {
         // verify
         assertNotPresent(app.listScopeConsentGrants(), app.getScopeConsentGrant(oAuth2ScopeConsentGrant.getId()))
     }
+
+    @Test
```
arvindkrishnakumar-okta, line 720: Add one test with `application/json` only.

VitaliiTytarenko-okta: ..or is it better to add a test without “Accept” at all? In this case “application/json” should be added in `applyDefaultRequestHeaders()`

arvindkrishnakumar-okta: yeah, that is fine.

api/pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <groupId>com.okta.sdk</groupId>
         <artifactId>okta-sdk-root</artifactId>
-        <version>3.0.4-SNAPSHOT</version>
+        <version>4.0.0-SNAPSHOT</version>
```
arvindkrishnakumar-okta, line 24: We should set the version to `3.1.0-SNAPSHOT` instead (`minor` version updgrade)

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -467,17 +483,28 @@ private Response execute(Request request) throws ResourceException {
 
         Assert.notNull(response, "response argument cannot be null.");
 
-        Map<String, Object> out = null;
+        if (response.hasBody() && (MediaType.APPLICATION_JSON.equals(response.getHeaders().getContentType())
+            || APPLICATION_ION_JSON.equals(response.getHeaders().getContentType()))) {
```
bdemers, line 487: nit: This is probably fine for now, given that Okta only has app/json and ion endpoints, but... `endsWith("+json")` might be more correct: https://tools.ietf.org/html/rfc6839

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -467,17 +483,28 @@ private Response execute(Request request) throws ResourceException {
...
+        return java.util.Collections.emptyMap();
+    }
+
+    private Optional<InputStream> getRawBody(Response response) {
+        Assert.notNull(response, "response argument cannot be null.");
 
         if (response.hasBody()) {
-            out = mapMarshaller.unmarshal(response.getBody(), response.getHeaders().getLinkMap());
+            return Optional.ofNullable(response.getBody());
         }
```
bdemers, line 499: this reads funny, if response has body, then Optional.ofNullable

Why not use `Optional.of(value)` instead, or skip the if?

arvindkrishnakumar-okta (approved):

impl/src/test/groovy/com/okta/sdk/impl/ds/DefaultDataStoreTest.groovy
```diff
@@ -89,11 +91,13 @@ class DefaultDataStoreTest {
         def responseBody = '{"name": "jcoder"}'
         def defaultDataStore = new DefaultDataStore(requestExecutor, "https://api.okta.com/v1", apiKeyResolver)
         def requestCapture = ArgumentCaptor.forClass(Request)
+        def headers = new HttpHeaders()
+        headers.setContentType(MediaType.APPLICATION_JSON)
```
bdemers, line 95: It looks like the existing tests _should_ sill function as is without these changes.  Ideally, we don't change existing tests.

If this line is removed does the test still pass?

VitaliiTytarenko-okta: No, they don't.
We need to set the "Accept" header for a response.

VitaliiTytarenko-okta: small clarification:
**DefaultDataStoreTest** is mocked, this is mean that the headers from the request will not be transferred to the response. That's why we need to set them manually to the `mock(Response)`;

arvindkrishnakumar-okta: @VitaliiTytarenko-okta Before this PR's change, the header used to be set by default for all requests at https://github.com/okta/okta-sdk-java/blob/master/impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java#L480

This is why the tests did not require explicit setting of headers earlier. Now with this PRs change, we should not be adding explicit tests in the Unit tests and it should work without it.

VitaliiTytarenko-okta: It doesn’t matter for those tests, because they are dummies. Even when the request contains the “Accept” header - the response doesn’t, because it doesn’t transfer it.
The only difference: earlier we did no do `response.getHeaders().getContentType()` (line 485 of ‘getBody’)
To return the correct response, it should now be: `MediaType.APPLICATION_JSON.equals(response.getHeaders().getContentType()) == true`
But previously was: 
```
when(response.getHeaders()).thenReturn(new HttpHeaders()) -> MediaType.APPLICATION_JSON.equals(null) == false
```

arvindkrishnakumar-okta (approved):

Issue #539, opened 2021-02-12
No enum constant com.okta.sdk.resource.application.ApplicationSignOnMode.RADIUS_FOR_GENERIC_APP


phi1ipp:
I'm trying to filter apps by their type with the snippet below and ran into the runtime exception (library version 3.0.2)
```
    client.listApplications()
        .filter { app -> app.signOnMode == ApplicationSignOnMode.SAML_2_0 }
        .forEach { app ->
        println(app)
    }
```

> Exception in thread "main" java.lang.IllegalArgumentException: No enum constant com.okta.sdk.resource.application.ApplicationSignOnMode.RADIUS_FOR_GENERIC_APP
	at java.base/java.lang.Enum.valueOf(Enum.java:273)
	at com.okta.sdk.impl.resource.EnumConverter.fromValue(EnumConverter.java:56)
	at com.okta.sdk.impl.resource.AbstractPropertyRetriever.getEnumProperty(AbstractPropertyRetriever.java:282)
	at com.okta.sdk.impl.resource.AbstractPropertyRetriever.getEnumProperty(AbstractPropertyRetriever.java:272)
	at com.okta.sdk.impl.resource.application.DefaultApplication.getSignOnMode(DefaultApplication.java:205)
	at MainKt.main(main.kt:25)
	at MainKt.main(main.kt)

arvindkrishnakumar-okta:
@phi1ipp Thanks for posting!

Presently, below is the exhaustive list of `ApplicationSignOnMode` values that the SDK understands:

https://github.com/okta/okta-management-openapi-spec/blob/master/resources/spec.yaml#L7288-L7298

I'll check to see if the spec is missing an update to include `RADIUS_FOR_GENERIC_APP` in `ApplicationSignOnMode` enum.

phi1ipp:
@arvindkrishnakumar-okta , thank you for a prompt response. I think it's missing in the enum as per my observation

arvindkrishnakumar-okta:
Internal Ref: OKTA-370307

Issue #184, opened 2018-04-27
Support unknown enum types


bdemers:
The backend service could extend any enum to add additional values.  This SDK needs a more graceful way of handling this (currently it would jus throw an exception)

arvindkrishnakumar-okta:
From 10.x onwards, the SDK uses Jackson to deserialize backend responses and the deserializer is configured to silently ignore unknown properties (and not fail). So, this issue is no longer applicable.

PR #544, opened 2021-02-18
Handle unknown Application SignOn Modes without error

sergiishamrai-okta:
## Issue(s)
https://github.com/okta/okta-sdk-java/issues/539

## Category

- [ ] Enhancement

### DIFF
--- a/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
+++ b/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
@@ -279,18 +279,15 @@ public abstract class AbstractPropertyRetriever implements PropertyRetriever {
 
         if (value != null) {
             if (value instanceof String) {
-                if(type.getName().equals("com.okta.sdk.resource.application.ApplicationSignOnMode")) {
-                    E convertedValue;
-                    try {
-                        convertedValue = enumConverter.fromValue(type, value.toString());
-                    } catch (IllegalArgumentException e) {
-                        convertedValue = enumConverter.fromValue(type, "CUSTOM");
-                        String msg = "Undeclared enum value. Defaulting to CUSTOM";
-                        log.error(msg, e);
-                    }
-                    return convertedValue;
+                E convertedValue;
+                try {
+                    convertedValue = enumConverter.fromValue(type, value.toString());
+                } catch (IllegalArgumentException e) {
+                    convertedValue = enumConverter.fromValue(type, "SDK_UNKNOWN");
+                    String msg = "Undeclared enum value. Defaulting to SDK_UNKNOWN";
+                    log.error(msg, e);
                 }
-                return enumConverter.fromValue(type, value.toString());
+                return convertedValue;
             }
             if (type.isAssignableFrom(value.getClass())) {
                 //noinspection unchecked
--- a/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
+++ b/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
@@ -320,11 +320,15 @@ public abstract class AbstractPropertyRetriever implements PropertyRetriever {
 
     protected void setProperty(Property property, Object value) {
         if(property.getType().isEnum() && value.toString().equals("SDK_UNKNOWN")) {
-            throw new IllegalArgumentException("The " + property.getType() + ".SDK_UNKNOWN value can not be used in setter");
+            throw new IllegalArgumentException(
+                "The " + property.getType() + ".SDK_UNKNOWN value can not be used in setter." +
+                    " Try to use put(\"" + property.getName() + "\", <your-value>)");
         } else if(property.getType().isEnum()
             && value instanceof List<?>
             && ((List) value).stream().anyMatch(x -> x.toString().equals("SDK_UNKNOWN"))) {
-            throw new IllegalArgumentException("The " + property.getType() + ".SDK_UNKNOWN value can not be used in setter");
+            throw new IllegalArgumentException(
+                "The " + property.getType() + ".SDK_UNKNOWN value can not be used in setter." +
+                    " Try to use put(\"" + property.getName() + "\", <your-value>)");
         }
         setProperty(property.getName(), value, true);
     }
--- a/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
+++ b/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
@@ -319,6 +319,13 @@ public abstract class AbstractPropertyRetriever implements PropertyRetriever {
     }
 
     protected void setProperty(Property property, Object value) {
+        if(property.getType().isEnum() && value.toString().equals("SDK_UNKNOWN")) {
+            throw new IllegalArgumentException("The " + property.getType() + ".SDK_UNKNOWN value can not be used in setter");
+        } else if(property.getType().isEnum()
+            && value instanceof List<?>
+            && ((List) value).stream().anyMatch(x -> x.toString().equals("SDK_UNKNOWN"))) {
+            throw new IllegalArgumentException("The " + property.getType() + ".SDK_UNKNOWN value can not be used in setter");
+        }
         setProperty(property.getName(), value, true);
     }
 
--- a/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
+++ b/impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
@@ -279,12 +279,18 @@ public abstract class AbstractPropertyRetriever implements PropertyRetriever {
 
         if (value != null) {
             if (value instanceof String) {
-                try {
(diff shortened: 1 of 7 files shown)


impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
```diff
@@ -279,7 +279,12 @@ protected Map getMapProperty(String key) {
 
         if (value != null) {
             if (value instanceof String) {
-                return enumConverter.fromValue(type, value.toString());
+                try {
```
arvindkrishnakumar-okta, line 282: I believe we should do more that just log the error and proceed. We should potentially generate an "UNKNOWN" type for all enums in the project and categorize the undocumented values received from backend into this "UNKNOWN" type. 

More specifically, for https://github.com/okta/okta-sdk-java/issues/539#issue-807402912 we should categorize them as `Custom`

sergiishamrai-okta: There is 108 enumerations in the project right now. I am not sure that adding another value into each enum is optimal solution. May be it's better to patch only one enum (ApplicationSignOnMode) by adding "Custom" value as [Python SDK](https://github.com/okta/okta-sdk-python/pull/168/files#diff-4fcfe22de0cfed07e17d41396cdf3bdb6d16863389c26595b2c184166263f25cR41) did?

arvindkrishnakumar-okta: Sounds reasonable to me.

arvindkrishnakumar-okta: Also, please add a test that covers this use case.

impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
```diff
@@ -279,6 +279,17 @@ protected Map getMapProperty(String key) {
...
+                    E convertedValue;
+                    try {
+                        convertedValue = enumConverter.fromValue(type, value.toString());
+                    } catch (IllegalArgumentException e) {
+                        convertedValue = enumConverter.fromValue(type, "CUSTOM");
+                        String msg = "Undeclared enum value. Defaulting to CUSTOM";
+                        log.error(msg, e);
+                    }
+                    return convertedValue;
+                }
```
bdemers, line 292: This block is going to need comments to explain what is going on here.
Especially with a note to remove it when: #184 is completed
use `ApplicationSignOnMode.class` instead of the string name

src/swagger/api.yaml
```diff
@@ -7346,6 +7346,7 @@ definitions:
       - SAML_2_0
       - OPENID_CONNECT
       - SAML_1_1
+      - CUSTOM
```
bdemers, line 7349: This needs to be in the official spec first:
https://github.com/okta/okta-management-openapi-spec/blob/master/dist/spec.yaml#L7323-L7336

impl/src/test/groovy/com/okta/sdk/impl/resource/AbstractPropertyRetrieverTest.groovy
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2021 Okta
```
bdemers, line 2:
```suggestion
 * Copyright 2021-Present Okta, Inc.
```

bdemers (changes requested):
Ideally, just add `RADIUS_FOR_GENERIC_APP` to the openapi spec, if it's a valid option, all the SDKs will have this problem (so fix it for all of them).

See #184 for a more generic longer-term solution, where the SDK no longer would use the `Enum` type, but use an enum-like class, that all have an `OTHER` option (probably not the actual string "OTHER", but used to illustrate the point)

I parsing fails, because a new type was added on the server side, the SDK returns the `OTHER` type, which has some sort of `getValue()` method, so the caller can retrieve the correct value.

Another short-term workaround is to use the `.get("key", String.class)` method which should skip all of the enum parsing, and return the raw value from the underlying JSON object.

sergiishamrai-okta:
@bdemers May be it will be much easier to catch exception right inside `getEnumProperty` method as it was presented [here](https://github.com/okta/okta-sdk-java/pull/544/files/5e2e8a726cdefe6806c52b05dd369df264d2bfcf) ? In this case we do not need to add a new property to enum, plus it will work for all enumerations in the SDK.

bdemers:
We would need to make sure there is a LOW chance of collision from the official API, so `UNKNOWN` probably isn't a great option, but something like `SDK_UNKNOWN` or some other name, could work.  There are some rough edges with the approach and probably a bit more special handling would be needed elsewhere in the code. 🤔 
* Any `toString` methods would need to show the original value
* a runtime error should be thrown if `SDK_UNKNOWN` value is set on a user, with some helpful message like use `object.set("key", "YOUR_VALUE")` instead

It does have the potential upside of being backward compatible and could be added into the enums via the templates (avoiding needing to change the openapi spec)

impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
bretterer: @bdemers We can't add this new app type to the spec, as its not documented. We are going to set it so a app type `signonmode` that doesn't exist in the mapping returns a `DefaultApplication` as we do in other SDK's

impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
```diff
@@ -279,7 +279,15 @@ protected Map getMapProperty(String key) {
...
         if (value != null) {
             if (value instanceof String) {
-                return enumConverter.fromValue(type, value.toString());
+                E convertedValue;
+                try {
+                    convertedValue = enumConverter.fromValue(type, value.toString());
+                } catch (IllegalArgumentException e) {
+                    convertedValue = enumConverter.fromValue(type, "SDK_UNKNOWN");
+                    String msg = "Undeclared enum value. Defaulting to SDK_UNKNOWN";
+                    log.error(msg, e);
```
bdemers, line 288: This isn't an "error" as it is recoverable, it should probably be a "warn", the error message should include the original string value as well, and probably some indication of how the user should access the value i.e. `.get("key", String.class)` (where `key` is the actual key string if possible

bdemers (commented):
This looks fine, assuming that's the approach @arvindkrishnakumar-okta is committing too.
I left a few comments before on the fact that we need to ensure that this new enum value is NOT allowed to be used in a setter (with instructions on what the user should do, i.e. call `.set("key", "new-value")` instead, this should be some type of RuntimeException (illegal arg or something), as this value would NOT be value if sent to the server, and the response would likely be less-obvious at that point.

arvindkrishnakumar-okta (approved):
lgtm

impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
```diff
@@ -279,7 +279,15 @@ protected Map getMapProperty(String key) {
 
         if (value != null) {
             if (value instanceof String) {
-                return enumConverter.fromValue(type, value.toString());
+                E convertedValue;
+                try {
+                    convertedValue = enumConverter.fromValue(type, value.toString());
+                } catch (IllegalArgumentException e) {
+                    convertedValue = enumConverter.fromValue(type, "SDK_UNKNOWN");
+                    String msg = "Undeclared enum value {}.{}. Defaulting to SDK_UNKNOWN";
```
bdemers, line 287: I think the `key` is important here. The warning needs to instruct the developer on what to do, i.e. call `"get(" + key + ", String.class)"` (or whatever the actual method is)

impl/src/main/java/com/okta/sdk/impl/resource/DefaultApplicationBuilder.java
```diff
@@ -65,6 +65,10 @@ public T setSelfService(Boolean selfService) {
 
     @Override
     public T setSignOnMode(ApplicationSignOnMode signOnMode) {
+        if(signOnMode == ApplicationSignOnMode.SDK_UNKNOWN) {
+            throw new IllegalArgumentException(
+                "The " + signOnMode.getClass().getName() + ".SDK_UNKNOWN can not be used in setter");
+        }
```
bdemers, line 71: If this logic wasn't here would the above `setProperty` logic get triggered?

sergiishamrai-okta: No. The `DefaultApplicationBuilder` class doesn't extend the abstract `AbstractPropertyRetriever` class. That is why we need this logic here.

impl/src/main/java/com/okta/sdk/impl/resource/AbstractPropertyRetriever.java
sergiishamrai-okta: @bdemers updated

arvindkrishnakumar-okta (approved):
lgtm

bdemers (approved):

Issue #239, opened 2018-12-10
ClientBuilder needs connection test


jamesdcarroll:
Just starting out with SDK so built simple client per doc.  No matter what orgURL or token I pass, however, there's no indication of success or failure. `client` is always not null and its not until I try to make a call that I get 401 for bad token or  `java.net.UnknownHostException` for bad address.  Seems like that should be a way to determine if the client is "ready"; either by throwing those errors from `build()` (or a `boolean isReaday` function that does) or setting `client` to null.

bdemers:
Hey @jamesdcarroll!

The client doesn't make any URL connections until you call a client method.  I do like the idea of an `isReady()` method or something like that.  That said these errors could also happen after the client is initialized (for example: if there was a network issue, or the API key was revoked)

Were you able to figure out what the issue was based on the descriptions in the exceptions/logs?

jamesdcarroll:
Yea, the funny thing is I knew the error going in as I was testing how it handled badURL or badToken so I could write the error handling code.  But build() would throw anything so it was irking me that I couldn't really validate that what I had gotten back from build() was usable before I tried to make a call.

bdemers:
We could do this a few ways 🤔 

```java
client.isReady()
```
```java
Clients.builder()
  .validateConnection()
  .build()
```

jamesdcarroll:
If it were me, I would vote for 

```
Clients.builder()
  .build()
  
client.isValid();
```
That way I can write 
```
Clients.builder()
  .build()
  if(client == null{
    for (Error err: client.errors){
         log.fatal("Connenction could not be built:" + err.message;
         return nulll;
     }
   }
 if (!client.isValid){
  for (Error err: client.errors){
         log.fatal("Connenction could not be validated:" + err.message;
         return client;
    }
   }
```

bdemers:
Trying to think about this a little more, in the case of an "invalid" client what would your application do? Throw an exception? Or go down another logic path?

VitaliiTytarenko-okta:
The new method `isReady` was added to the DataStore.
see: https://github.com/okta/okta-sdk-java/blob/e02a77d891fffa7e44e33da44263654a765e28eb/api/src/main/java/com/okta/sdk/ds/DataStore.java#L124

PR #571, opened 2021-04-02
add Client connection test

VitaliiTytarenko-okta:
## Issue(s)
https://github.com/okta/okta-sdk-java/issues/239

## Description
Add `isReady()` method to DataStore for checking if **okta.client.orgUrl** and **okta.client.token** are correct, Okta server is available and DataStore can execute requests. Accept a reference to the method based on which the ready state will be checked.

```
    Client client = Clients.builder().build();
    boolean isReadyState = client.isReady(client::listApplications);      // <true> if configured correctly
```

## Category
- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,14 @@ Client client = Clients.builder()
  
 Hard-coding the Okta domain and API token works for quick tests, but for real projects you should use a more secure way of storing these values (such as environment variables). This library supports a few different configuration sources, covered in the [configuration reference](#configuration-reference) section.
 
+In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It doesn't produce exceptions if the wrong orgUrl or token have been used, but it returns boolean. As a parameter, you have to provide a method reference based on which the ready state will be checked. Recommended realization:
+
+[//]: # (method: isReadyChecking)
+```java
+boolean isClientReadyStatus = client.isReady(client::listApplications);
+```
+[//]: # (end: isReadyChecking)
+
 ## OAuth 2.0
 
 Okta allows you to interact with Okta APIs using scoped OAuth 2.0 access tokens. Each access token enables the bearer to perform specific actions on specific Okta endpoints, with that ability controlled by which scopes the access token contains.
@@ -135,10 +143,10 @@ Client client = Clients.builder()
     .setClientId("{clientId}")
     .setScopes(new HashSet<>(Arrays.asList("okta.users.read", "okta.apps.read")))
     .setPrivateKey("/path/to/yourPrivateKey.pem")
-    // (or) .setPrivateKey("full PEM payload")
-    // (or) .setPrivateKey(Paths.get("/path/to/yourPrivateKey.pem"))
-    // (or) .setPrivateKey(inputStream)
-    // (or) .setPrivateKey(privateKey)
+    // (or) .setPrivateKey("full PEM payload");
+    // (or) .setPrivateKey(Paths.get("/path/to/yourPrivateKey.pem"));
+    // (or) .setPrivateKey(inputStream);
+    // (or) .setPrivateKey(privateKey);
     .build();
 ```
 [//]: # (end: createOAuth2Client)
--- a/README.md
+++ b/README.md
@@ -40,10 +40,11 @@ This library uses semantic versioning and follows Okta's [library version policy
 | Version | Status                    |
 | ------- | ------------------------- |
 | 0.0.x | :warning: Retired |
-| 1.x   | :clock9: Retiring effective April 07, 2021 |
+| 1.x   | :warning: Retired |
 | 2.x.x | :clock9: Retiring effective September 03, 2021 |
-| 3.x.x | :heavy_check_mark: Stable ([migration guide](https://github.com/okta/okta-sdk-java/blob/master/MIGRATING.md#migrating-from-2xx-to-300)) |
- 
+| 3.x.x | :clock9: Retiring effective November 10, 2021 |
+| 4.x.x | :heavy_check_mark: Stable ([migration guide](https://github.com/okta/okta-sdk-java/blob/master/MIGRATING.md#migrating-from-3xx-to-400)) |
+
 The latest release can always be found on the [releases page][github-releases].
  
 ## Need help?
--- a/README.md
+++ b/README.md
@@ -144,10 +144,10 @@ Client client = Clients.builder()
     .setClientId("{clientId}")
     .setScopes(new HashSet<>(Arrays.asList("okta.users.read", "okta.apps.read")))
     .setPrivateKey("/path/to/yourPrivateKey.pem")
-    // (or) .setPrivateKey("full PEM payload");
-    // (or) .setPrivateKey(Paths.get("/path/to/yourPrivateKey.pem"));
-    // (or) .setPrivateKey(inputStream);
-    // (or) .setPrivateKey(privateKey);
+    // (or) .setPrivateKey("full PEM payload")
+    // (or) .setPrivateKey(Paths.get("/path/to/yourPrivateKey.pem"))
+    // (or) .setPrivateKey(inputStream)
+    // (or) .setPrivateKey(privateKey)
     .build();
 ```
 [//]: # (end: createOAuth2Client)
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Hard-coding the Okta domain and API token works for quick tests, but for real pr
 
 In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It does not produce exceptions if the wrong orgUrl or token have been used, but it returns a boolean indicating the client readiness.
 
-[//]: # (method: isReadyChecking)
+[//]: # (method: isClientReady)
 ```java
 boolean isClientReadyStatus = client.isReady(client::listApplications);
 ```
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Client client = Clients.builder()
  
 Hard-coding the Okta domain and API token works for quick tests, but for real projects you should use a more secure way of storing these values (such as environment variables). This library supports a few different configuration sources, covered in the [configuration reference](#configuration-reference) section.
 
-In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It doesn't produce exceptions if the wrong orgUrl or token have been used, but it returns boolean. As a parameter, you have to provide a method reference based on which the ready state will be checked. Recommended realization:
+In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It does not produce exceptions if the wrong orgUrl or token have been used, but it returns a boolean indicating the client readiness.
 
 [//]: # (method: isReadyChecking)
 ```java
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ In some cases, it maybe needed to check if the client is ready and able to execu
 ```java
 boolean isClientReadyStatus = client.isReady(client::listApplications);
 ```
-[//]: # (end: isReadyChecking)
+[//]: # (end: isClientReady)
 
 ## OAuth 2.0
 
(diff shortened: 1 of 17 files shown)


pom.xml
```diff
@@ -343,8 +343,8 @@
                         </oldVersion>
                         <parameter>
                             <onlyModified>true</onlyModified>
-                            <breakBuildOnBinaryIncompatibleModifications>true</breakBuildOnBinaryIncompatibleModifications>
-                            <breakBuildBasedOnSemanticVersioning>true</breakBuildBasedOnSemanticVersioning>
+                            <breakBuildOnBinaryIncompatibleModifications>false</breakBuildOnBinaryIncompatibleModifications>
+                            <breakBuildBasedOnSemanticVersioning>false</breakBuildBasedOnSemanticVersioning>
```
bdemers, line 347: You don't want to disable this

integration-tests/src/test/groovy/com/okta/sdk/tests/it/ApplicationsIT.groovy
```diff
@@ -99,6 +99,11 @@ class ApplicationsIT extends ITSupport {
         }
     }
 
+    @Test
+    void testClientIsReady() {
+        assertThat client.isReady(), is(true)
```
bdemers, line 104: Avoid using `is(true)` or `is(false)` it does not provide a reason when it fails, instead set the reason manually:
```suggestion
        assertThat "Expected client to be ready", client.isReady()
```

Or something like that

impl/src/test/groovy/com/okta/sdk/impl/ds/DefaultDataStoreTest.groovy
```diff
@@ -229,4 +230,30 @@ class DefaultDataStoreTest {
...
+    @Test
+    void testIsReadyTrue(){
+        def response = mock(Response)
+        def requestExecutor = mock(RequestExecutor)
+        def clientCredentialsResolver = mock(ClientCredentialsResolver)
+
+        when(requestExecutor.executeRequest(any())).thenReturn(response)
+        when(response.getBody()).thenReturn(new StringInputStream('{"name": "web app"}'))
+
+        def defaultDataStore = new DefaultDataStore(requestExecutor, "https://api.okta.com/v1", clientCredentialsResolver)
```
bdemers, line 243:
```suggestion
        def defaultDataStore = new DefaultDataStore(requestExecutor, "https://api.example.com/v1", clientCredentialsResolver)
```

impl/src/test/groovy/com/okta/sdk/impl/ds/DefaultDataStoreTest.groovy
```diff
@@ -229,4 +230,30 @@ class DefaultDataStoreTest {
...
+    }
+
+    @Test
+    void testIsReadyFalse(){
+        def requestExecutor = mock(RequestExecutor)
+        def clientCredentialsResolver = mock(ClientCredentialsResolver)
+
+        when(requestExecutor.executeRequest(any())).thenThrow(new HttpException("Unable to execute HTTP request"))
+
+        def defaultDataStore = new DefaultDataStore(requestExecutor, "https://api.okta.com/v1", clientCredentialsResolver)
```
bdemers, line 255:
```suggestion
        def defaultDataStore = new DefaultDataStore(requestExecutor, "https://api.example.com/v1", clientCredentialsResolver)
```

impl/src/test/groovy/com/okta/sdk/impl/ds/DefaultDataStoreTest.groovy
bdemers: use `example.com` when possible in UT's

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -595,4 +596,13 @@ private HttpHeaders headers(Map<String, List<String>> headerParameters) {
     public RequestBuilder http() {
         return new DefaultRequestBuilder(this);
     }
+
+    @Override
+    public boolean isReady() {
+        try {
+            return getRawResponse("/api/v1/apps", null, null) != null;
+        } catch (ResourceException | HttpException exception) {
+            return false;
```
bdemers, line 605: this exception should be logged, probably at debug.

I'd also question the use of the apps API here, In the case of OAuth tokens, the token might not be scoped to be able to use this API, which may cause a false positive. cc: @arvindkrishnakumar-okta

bdemers: In that case it would be difficult to troubleshoot as the exception isn't logged ;)

bdemers (changes requested):

api/src/main/java/com/okta/sdk/ds/DataStore.java
```diff
@@ -113,4 +113,11 @@
      */
     RequestBuilder http();
 
+    /**
+     * Check if Datastore is configured correctly and able to execute requests.
+     *
+     * @return {@code true} if orgUrl and token are correct and Okta server is available, otherwise {@code false}.
```
arvindkrishnakumar-okta, line 119: Nit: expand a bit on `Okta server is available`.

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
arvindkrishnakumar-okta: I guess the need here is to just ensure that the connectivity between client and Okta Org is fine. Assuming this, a simple `ping` functionality should suffice here instead of a fully qualified API URL to hit?

pom.xml
VitaliiTytarenko-okta: I did it for a successful build. I'll revert it later

impl/src/main/java/com/okta/sdk/impl/ds/DefaultDataStore.java
arvindkrishnakumar-okta: I think the need here is to just ensure that the connectivity between client and Okta Org is fine. We could probably hit the OIDC discovery endpoint `https://{yourOktaDomain}/.well-known/openid-configuration` and determine if the client-server communication channel is ready.

bdemers: if the goal is to see that the client is ready (and can authenticate). The well-known endpoint might not work as it's anonymous.

I'm not sure this feature could reliably work for OAuth clients (given the potential scope issues).  You can also scope an SSWS token, so there are some potential issues there too 🤔 

Maybe this is a feature we recommend folks implement on their own? and provide some sort of doc?
basically what @VitaliiTytarenko-okta originally had, but in a README instead?

```java
 try {
    client.listApplications()
} catch (ResourceException exception) {
    // handle client not ready state
}
```

It's ugly, or we could also do something with a lambda:

```java
boolean ready = client.isReady(() -> client.listApplications())
```

I don't think it's obvious what that does, but.. we could also take the highbred approach and default `isReady()` to the apps endpoint as well.
While a bit more elegant than the first, it comes with the cost of complicated docs. It's also not something a client _should_ regularly do except in the case of some sort to initialization loop, which might already be custom logic?  🤔   

Thoughts?

arvindkrishnakumar-okta: We can possibly do something like `isReady("client-specified-url)"` instead of a preset `list apps` URL?

VitaliiTytarenko-okta: Unfortunately, I haven't come up with anything better. I don't like an idea to pass URL as a parameter, because any URL can be passed and we can't guarantee anything. So I used a method reference as a `Supplier<? extends Resource>`.

api/src/main/java/com/okta/sdk/ds/DataStore.java
VitaliiTytarenko-okta: Here I meant not a particular server but Okta Core in general and _available_ means "able to be used or obtained". I think everything is quite clear here. Please, feel free to post your variant of the description.

README.md
```diff
@@ -120,6 +120,14 @@ Client client = Clients.builder()
  
 Hard-coding the Okta domain and API token works for quick tests, but for real projects you should use a more secure way of storing these values (such as environment variables). This library supports a few different configuration sources, covered in the [configuration reference](#configuration-reference) section.
 
+In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It doesn't produce exceptions if the wrong orgUrl or token have been used, but it returns boolean. As a parameter, you have to provide a method reference based on which the ready state will be checked. Recommended realization:
```
arvindkrishnakumar-okta, line 123:
```suggestion
In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It does not produce exceptions if the wrong orgUrl or token have been used, but it returns a boolean indicating the client readiness.
```

examples/quickstart/src/main/java/quickstart/ReadmeSnippets.java
```diff
@@ -67,6 +67,10 @@ private void createClient() {
             .build();
     }
 
+    private void isReadyChecking() {
```
arvindkrishnakumar-okta, line 70:
```suggestion
    private void isClientReady() {
```

README.md
```diff
@@ -120,6 +120,14 @@ Client client = Clients.builder()
  
 Hard-coding the Okta domain and API token works for quick tests, but for real projects you should use a more secure way of storing these values (such as environment variables). This library supports a few different configuration sources, covered in the [configuration reference](#configuration-reference) section.
 
+In some cases, it maybe needed to check if the client is ready and able to execute requests. The _**isReady**_ method can be used for this. It does not produce exceptions if the wrong orgUrl or token have been used, but it returns a boolean indicating the client readiness.
+
+[//]: # (method: isReadyChecking)
```
arvindkrishnakumar-okta, line 125:
```suggestion
[//]: # (method: isClientReady)
```

README.md
```diff
@@ -120,6 +120,14 @@ Client client = Clients.builder()
...
+```java
+boolean isClientReadyStatus = client.isReady(client::listApplications);
+```
+[//]: # (end: isReadyChecking)
```
arvindkrishnakumar-okta, line 129:
```suggestion
[//]: # (end: isClientReady)
```

VitaliiTytarenko-okta:
@bdemers @arvindkrishnakumar-okta 
Can this be merged?

README.md
```diff
@@ -136,10 +144,10 @@ Client client = Clients.builder()
     .setClientId("{clientId}")
     .setScopes(new HashSet<>(Arrays.asList("okta.users.read", "okta.apps.read")))
     .setPrivateKey("/path/to/yourPrivateKey.pem")
-    // (or) .setPrivateKey("full PEM payload")
-    // (or) .setPrivateKey(Paths.get("/path/to/yourPrivateKey.pem"))
-    // (or) .setPrivateKey(inputStream)
-    // (or) .setPrivateKey(privateKey)
+    // (or) .setPrivateKey("full PEM payload");
```
sergiishamrai-okta, line 147: Do we really need a semicolon symbol here?

arvindkrishnakumar-okta:
> @bdemers @arvindkrishnakumar-okta
> Can this be merged?

LGTM and ready to merge once all the minor outstanding comments are addressed.

README.md
VitaliiTytarenko-okta: 🤣 
@sergiishamrai-okta It is your changes from here: https://github.com/okta/okta-sdk-java/blob/4170c6847e1ecbc88639a23ee9c424c87af67920/examples/quickstart/src/main/java/quickstart/ReadmeSnippets.java#L80

arvindkrishnakumar-okta (approved):

pom.xml
arvindkrishnakumar-okta: Please create a followup JIRA item for this so we dont forget.

PR #604, opened 2021-07-19
OKTA-347728: Fetch and use Open API spec from npm

VitaliiTytarenko-okta:
## Story
[OKTA-347728](https://oktainc.atlassian.net/browse/OKTA-347728)

## Description
Presently the Java Mgmt SDK is using a local copy of the Open API spec. It was originally designed that way and should be changed to use the one from NPM.

We should update the SDK to fetch the Open API spec from npm package @okta/openapi directly like other management SDKs do today.

Note: Presently, the openapi repo is private and is in the process of getting moved to public.

Use VM option: `-Dexec.skip=true` to skip this fetching.

## Category
- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature
- [x] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation


arvindkrishnakumar-okta (changes requested):
You are missing license file headers, see build error.

VitaliiTytarenko-okta:
@arvindkrishnakumar-okta 
The build failed because there is no license header in `/node_modules/@okta/openapi/dist/spec.yaml`. We need to include it in the OpenApi repository.
Also, we have to publish the current version 2.5.0 of OpenApi to `npmjs`.

arvindkrishnakumar-okta:
> /node_modules/@okta/openapi/dist/spec.yaml

@VitaliiTytarenko-okta Please open a PR in https://github.com/okta/okta-management-openapi-spec repo to incude license header. Once you're done, lets wait for the next NPM release.

VitaliiTytarenko-okta:
[OpenApi PR](https://github.com/okta/okta-management-openapi-spec/pull/51)

arvindkrishnakumar-okta (approved):

examples/quickstart/pom.xml
```diff
@@ -60,8 +60,17 @@
...
                 <artifactId>exec-maven-plugin</artifactId>
-                <version>1.3.2</version>
                 <executions>
+                    <execution>
+                        <id>copy to swagger</id>
+                        <configuration>
+                            <executable>sh</executable>
+                            <arguments>
+                                <argument>-c</argument>
+                                <argument>cp ../../node_modules/@okta/openapi/dist/spec.yaml ../../src/swagger/api.yaml</argument>
```
arvindkrishnakumar-okta, line 70: @VitaliiTytarenko-okta why is this plugin config needed in examples/quickstart pom.xml?

pom.xml
```diff
@@ -246,6 +246,40 @@
     <build>
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>exec-maven-plugin</artifactId>
```
arvindkrishnakumar-okta, line 251: @VitaliiTytarenko-okta can we also add config to skip this plugin execution when needed? Ref: https://www.mojohaus.org/exec-maven-plugin/exec-mojo.html#skip

I am thinking abt cases where we need to skip this execution (in case of any issues with the spec from npm) and use the local spec (src/swagger/api.yaml) in this case. I think we should add this capability.

arvindkrishnakumar-okta (changes requested):
see comments

examples/quickstart/pom.xml
VitaliiTytarenko-okta: In fact, it is not necessary, but I had to add this part here because `exec-maven-plugin` was already defined in this module. Therefore without these lines, I had an error during building due to a different package location.
But `<skip>` option can solve this problem.

pom.xml
VitaliiTytarenko-okta: of course

arvindkrishnakumar-okta:
If we use the `-Dexec.skip=true` while building, would the SDK still use the local copy in `src/swagger/api.yaml`? This means we have to maintain the copy at `src/swagger/api.yaml` current with the latest spec [here](https://github.com/okta/okta-management-openapi-spec/blob/master/resources/spec.yaml).

impl/pom.xml
```diff
@@ -151,6 +151,10 @@
         </resources>
 
         <plugins>
+            <plugin>
```
arvindkrishnakumar-okta, line 154: Is this plugin needed here?

VitaliiTytarenko-okta:
> If we use the `-Dexec.skip=true` while building, would the SDK still use the local copy in `src/swagger/api.yaml`? This means we have to maintain the copy at `src/swagger/api.yaml` current with the latest spec [here](https://github.com/okta/okta-management-openapi-spec/blob/master/resources/spec.yaml).

@arvindkrishnakumar-okta 
And introduce one more optional flag for cases when we don't want to update `api.yaml` at all? I think that this is an overabundance and we shouldn't even touch `api.yaml` if the skip option is specified.
Do we really need to maintain an intermediate version of `api.yaml` between releases?

impl/pom.xml
VitaliiTytarenko-okta: no need

api/pom.xml
```diff
@@ -59,6 +59,10 @@
 
     <build>
         <plugins>
+            <plugin>
```
arvindkrishnakumar-okta, line 62: Is this plugin config needed here?

src/swagger/api.yaml
```diff
@@ -13508,6 +13508,12 @@ definitions:
         type: array
       pattern:
         type: string
+      oneOf:
```
arvindkrishnakumar-okta, line 13511: @VitaliiTytarenko-okta if the spec is fetched from npm, why do we need to do this update to the spec's local copy? I am not sure if I am missing something here.

api/pom.xml
VitaliiTytarenko-okta: Yes.
We don't need it in both modules: `api` and `impl`. Because it is enough for us to have it executed only once.
And we should it have here (but not in `impl`) as `api` module will be built before `impl` https://github.com/okta/okta-sdk-java/blob/dc7ac324de111c8f71a22c3e53023d5c9cf973d4/pom.xml#L52

src/swagger/api.yaml
VitaliiTytarenko-okta: Maybe completely exclude this file from this repository (.gitignore)?
When the command `mvn compile` is executed, **api.yaml** will be generated locally.

arvindkrishnakumar-okta: The npm spec will be copied/overwritten to `src/swagger/api.yaml` right?

VitaliiTytarenko-okta: Yes

arvindkrishnakumar-okta: Why would we update this manually (as shown here) then?

VitaliiTytarenko-okta: Updated

pom.xml
```diff
@@ -343,8 +379,8 @@
                         </oldVersion>
                         <parameter>
                             <onlyModified>true</onlyModified>
-                            <breakBuildOnBinaryIncompatibleModifications>true</breakBuildOnBinaryIncompatibleModifications>
-                            <breakBuildBasedOnSemanticVersioning>true</breakBuildBasedOnSemanticVersioning>
+                            <breakBuildOnBinaryIncompatibleModifications>false</breakBuildOnBinaryIncompatibleModifications>
```
arvindkrishnakumar-okta, line 382: Both these flags should be set to `true`. 

Note that `v5.0.0` has already been released with breaking changes.

arvindkrishnakumar-okta (approved):

PR #649, opened 2021-10-02
Add a BOM for easier dependency management

jamietanna:
## Issue(s)

N/A

## Description

To allow for easier managing of dependencies for consumers of the Okta
SDK, we can introduce a Bill of Materials (BOM).

This allows managing dependencies easier, and means bumping all the dependency versions can be done at the same time.

## Category

- [ ] Bugfix
- [x] Enhancement
- [x] New Feature
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit Test(s)
- [ ] Documentation

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2021 Okta
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.okta.sdk</groupId>
+        <artifactId>okta-sdk-root</artifactId>
+        <version>6.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>okta-sdk-bom</artifactId>
+    <name>Okta Java SDK :: BOM</name>
+    <description>
+        Bill of Materials pom for getting full, complete set of compatible versions of Okta SDK components.
+    </description>
+    <packaging>pom</packaging>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.okta.sdk</groupId>
+                <artifactId>okta-sdk-sswagger-templates</artifactId>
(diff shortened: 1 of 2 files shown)


sergiishamrai-okta:
Hi @jamietanna,
Thanks for the PR.
We'll investigate the necessity of such an approach and back to you.

arvindkrishnakumar-okta:
Thanks for the PR.

Request you to please sign the CLA at https://developer.okta.com/cla/

bdemers (changes requested):
Thanks @jamietanna!

We probably want to tweak this a little, probably remove the test, examples, and coverage modules.  Only the `api`, `impl`, and HTTP client libs _should_ be needed.  I usually recommend against using `project.version` (to avoid a few edge cases).

Taking a step back though, if we create a BOM, it _might_ be better to create a separate project and include all other related Okta libs, (SDK, Auth libs, JWT verifier, etc).

@jamietanna, I'm guessing you are using this SDK with another Okta lib and need to make sure the versions are aligned?  Or are you seeing version/dependency issues with other projects?

jamietanna:
Hey @bdemers thanks for the comments!

> Only the api, impl, and HTTP client libs should be needed.

Sure thing, I'll amend that :+1:

>  I usually recommend against using project.version (to avoid a few edge cases).

Interesting, what would be the alternative you'd recommend?

> Taking a step back though, if we create a BOM, it might be better to create a separate project and include all other related Okta libs, (SDK, Auth libs, JWT verifier, etc).

That's a very fair shout! Would it make sense to be here, or a completely fresh GitHub repo + standalone project?

>  I usually recommend against using project.version (to avoid a few edge cases).

On one of [my side projects](https://gitlab.com/jamietanna/www-api/) I've been using [WhiteSource Renovate](https://www.whitesourcesoftware.com/free-developer-tools/renovate/) and have found that it's raised a few MRs:

- https://gitlab.com/jamietanna/www-api/-/merge_requests/293
- https://gitlab.com/jamietanna/www-api/-/merge_requests/294
- https://gitlab.com/jamietanna/www-api/-/merge_requests/295

This isn't a problem, but it'd be even better if it could bump them all together through a BOM!

bdemers:
I'm guessing it would need to be a separate GitHub repo just to avoid circular dependencies, but maybe there is an alternative we can come up with.

As for the `project.version` usage. IIRC and this was from Maven 2 days, (so it's possible these issues have been resolved).  There are a few cases where `project.version` can be resolved to a different version than was originally intended, for example, if I created a project that extended from this bom AND used a new version. `project.version` could resolve to the new version.  

Again, it's been years since tracking that type of issue down, so i could be miss remembering the details too.  We use the Versions plugin (`versions:set -DnewVersion=<new-version-here>`) and it takes care of all of updating the nested versions too.

Also, the chances of someone using this as parent pom are very low, so I'm probably just a bit paranoid 🤣

jamietanna:
Thanks, I've amended the `project.version` :+1:

Thinking about it, I still feel like this should be added on its own - as well as having i.e. an `okta-platform:okta-platform-bom` which includes all of the Okta projects, having an easy way to just update the clients together makes sense, especially as we've got a monorepo that is versioned together - thoughts?

jamietanna:
This certainly does work (and the Gradle equivalents) but though this would be a quick way to solve it for consumers.

I'd be happy to get a PR raised to a repo if you can create one with an initial commit for me?

(I was hoping this would be a Hacktoberfest contribution, but happy to wait if it's gonna take a bit of time to get the thinking / versioning strategy right for everything)

jamietanna:
Hey folks, any thoughts as to whether we'd be happy creating a BOM for this library on its own, before we look at a wider one?

sergiishamrai-okta:
Hi @jamietanna ,
Thanks for PR. We'll review it shortly.

sergiishamrai-okta:
A “bom” project could be created, but that would need to include all of our Java libraries (and live in a separate git repo), so closing this PR for now and waiting for a decision about creating "bom" project with all Java libraries.

PR #679, opened 2021-12-10
upload file operations

sergiishamrai-okta:
## Issue(s)
[OKTA-448365](https://oktainc.atlassian.net/browse/OKTA-448365)

## Description
Multipart/form-data support added to support file uploading.
ITs for `uploadBrandThemeLogo`, `deleteBrandThemeLogo`, `updateBrandThemeFavicon`, `deleteBrandThemeFavicon`, `updateBrandThemeBackgroundImage`, `deleteBrandThemeBackgroundImage` added.


sergiishamrai-okta:
Blocked by: https://github.com/okta/okta-commons-java/pull/66

impl/src/main/java/com/okta/sdk/impl/resource/DefaultFileResource.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.File;
+import java.util.Collections;
+import java.util.Map;
+
+public class DefaultFileResource extends AbstractResource implements FileResource {
+
+    private final File file;
+    private final String formDataName;
+
+    public DefaultFileResource(InternalDataStore dataStore, File file, String formDataName) {
```
bdemers, line 30: Requiring the use of a "File" limits the use cases where this can be used.  A stream should be used.

For example, an application could be loading the image from a database, or another remote location, and would not have a "File" directly. 

Supporting files directly is still a good idea too, but a `Path` should be used instead of a `File`.

bdemers (changes requested):

impl/src/main/java/com/okta/sdk/impl/resource/DefaultFileResource.java
sergiishamrai-okta: Updated. now `java.nio.file.Path` used instead of `java.io.File`

arvindkrishnakumar-okta (approved):

PR #731, opened 2022-04-24
Improve performance in the Test phase

CAIWan1998:
Maven will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones. We can fork more test VM by setting `<fork>1.5C</fork>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -300,7 +300,7 @@
                     <configuration>
                         <trimStackTrace>false</trimStackTrace>
                         <reuseForks>true</reuseForks>
-                        <forkCount>1</forkCount>
+                        <forkCount>1.5C</forkCount>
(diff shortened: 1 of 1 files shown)


arvindkrishnakumar-okta:
@CAIWan1998 Thanks for the PR!

While this change could prove efficient in terms of memory and execution time, it potentially makes the tests more vulnerable towards race conditions and hard to reproduce/inconsistent behavior.

CAIWan1998:
Hi @arvindkrishnakumar-okta !
As long as we write reasonable test cases, I belive this wouldn't  happen. As you said, this PR is designed to improve Build Performance.

PR #792, opened 2022-12-19
updated parallelism changes

karthikaselvaraj-okta:


.circleci/config.yml
```diff
@@ -5,7 +5,13 @@ aliases:
   - &build_steps
     - checkout
     - run: java -version
-    - run: ./mvnw clean install -Pci
+    - run: mvn clean install -Dmaven.test.skip=true
```
arvindkrishnakumar-okta, line 8: This configuration would actually skip the Unit tests also. We should not skip the UTs.

.circleci/config.yml
```diff
@@ -20,25 +26,31 @@ jobs:
       - image: cimg/openjdk:8.0.345-node
     environment:
       JVM_OPTS: -Xmx3200m
+    resource_class: large
+    parallelism: 3
```
arvindkrishnakumar-okta, line 30: What does this setting `3` do?

karthikaselvaraj-okta: Parallelism:3 --> It creates 3 different containers to run the IT in parallel

arvindkrishnakumar-okta (changes requested):
CircleCI build logs for Java 8, 11 & 17 all show that only 1 IT (IdpIT) is run for the below command:

```
    - run:
        command: |
          ClassName=$(circleci tests glob "integration-tests/src/test/groovy/com/okta/sdk/tests/it/**IT.groovy" | circleci tests split --split-by=filesize | grep groovy | xargs -n 1 basename | sed -e 's/\.groovy$//' | tr '\n' ',' | sed 's/,*$//g')
          echo $ClassName
          cd ./integration-tests
          mvn -Dtest=$ClassName test
```

image

vijetmahabaleshwar-okta (approved):

arvindkrishnakumar-okta (approved):

PR #948, opened 2023-09-11
Rework polymorphic deserialization to use built-in support in OpenAPI generator

clementdenis:
## Description

This PR changes the OpenAPI client generation to include polymorphic serialization annotations.
It seems these were missing from the custom pojo.mustache template (don't know if it was on purpose or from an older version).
I just copied the code from the latest version in the OpenAPI generator:
https://github.com/OpenAPITools/openapi-generator/blob/8b15d4820ed1ee206e2aab99e8f55dc27bed3e82/modules/openapi-generator/src/main/resources/Java/pojo.mustache#L28

With these changes, the custom deserializers, *ApiHelper classes and HelperUtil are not useful anymore.
I only kept the custom UserProfile (de)serializer, that is not related to polymorphism.
Also, all polymorphic models are now properly support, not only those with a specific deserializer.

The API is know a bit less type-safe, as there isn't the type-aware versions of the create / replace methods in *ApiHelper classes (developers need to cast the responses). But it seems like an acceptable trade-off given the reduced complexity of the SDK (and it was required anyway when getting objects).

I tested this on a few types like Application and LogStream, but this should work for all polymorphic models as the approach is generic and driven by the OpenAPI spec.

## Category
- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature
- [ ] Library Upgrade
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit or Integration Test(s) 
- [ ] Documentation

## Signoff
- [x] I have submitted a CLA for this PR
- [x] I have updated documentation to explain what my PR does


arvindkrishnakumar-okta:
@clementdenis Thanks for the PR, I'll review and merge this soon.

PR #980, opened 2024-02-07
Fix deserialization of polymorphic types when discriminator property is unknown or null

clementdenis:
## Description

Starting with version 14, polymorphic deserialization is handled by the default OpenAPI generator mechanism.

However, for some polymorphic types , the discriminator property can be null.
For example, Directory Integration apps like Active Directory have a null signOnMode, despite [the documentation claiming it shouldn't be](https://developer.okta.com/docs/reference/api/apps/#application-properties).

In that case, no error is triggered, because the Jackson ObjectMapper disables DeserializationFeature.FAIL_ON_INVALID_SUBTYPE.
But the resulting deserialized object is null, possibly causing NPEs in the client code.
This can also happen when the API returns an unknown value for the discriminator property.

To fix the problem, the Jackson annotations in the [typeInfoAnnotation.mustache](https://github.com/OpenAPITools/openapi-generator/blob/v6.6.0/modules/openapi-generator/src/main/resources/Java/typeInfoAnnotation.mustache) template have been tuned:
- Add `defaultImpl` to `@JsonTypeInfo` to deserialize to base type if discriminator is null or unknown
- Remove `@JsonIgnoreProperties` on discriminator field and set `@JsonTypeInfo.include=JsonTypeInfo.As.EXISTING_PROPERTY` (required to avoid the Application base class instances to have signOnMode="Application")
- This last part creates a small difference with current code: when instantiating subclasses of Application (or other polymorphic classes), the discriminator field must be set explicitly, or it will stay null.

I didn't add any test, as I'm not sure how to execute the AppsIT class, but checked serialization and deserialization is correct.

This PR also removes leftover Mixins used for deserialization before v14 and corresponding test class.

## Category
- [x] Bugfix
- [ ] Enhancement
- [ ] New Feature
- [ ] Library Upgrade
- [ ] Configuration Change
- [ ] Versioning Change
- [ ] Unit or Integration Test(s) 
- [ ] Documentation

## Signoff
- [x] I have submitted a CLA for this PR
- [x] I have updated documentation to explain what my PR does


clementdenis:
This raises another major issue with the current SDK: when deserializing a Directory Integration app like Active Directory (those with name=active_directory) to Application base class, some important fields are not mapped like name, credentials and settings

I suppose this specific case requires changing the API itself, to add a signOnMode value for those apps, and add the corresponding specialized application model in the OpenAPI spec.

A note on previous versions:
- Previous deserialization code in v13 did not handle that case either, as null and unknown signOnMode also defaulted to Application.
- v8 did not have this issue, as all API resources were extending ExtensibleResource, so they could hold arbitrary unknown field values.

But more generally, the API resource models should be able to hold unknown field values like on v8 => this does not easy to do with OpenAPI generator. `disallowAdditionalPropertiesIfNotPresent` ([from Java config options](https://openapi-generator.tech/docs/generators/java/#config-options)) seems to be what's needed, but it's not implemented on all libraries.

arvind15683:
@clementdenis Thanks for the PR!

I'll review it soon and get it merged.

arvindkrishnakumar-okta:
@clementdenis Thanks for the PR!

I'll get it reviewed and merged soon.

arvindkrishnakumar-okta:
Merging into a feature branch (that CCI can build) for testing before it gets finally merged to master/main branch.

clementdenis:
Hi @arvindkrishnakumar-okta,

Any chance this makes it into master?

arvindkrishnakumar-okta:
@clementdenis Sure, I'm seeing some CI issues, will get those sorted out and merge it.

PR #1700, opened 2026-07-21
fix: DomainValidationStatus.FAILED_TO_VERIFY, DevicePostureChecks.include typing, ApplicationGrantsIT CI fix

prachi-okta:
## Issue(s)

- OKTA-1226428 - `DomainValidationStatus` missing `FAILED_TO_VERIFY` enum constant
- OKTA-1221543 - `DevicePostureChecks.include` typed `List<String>` instead of a proper structured type

### 1. `DomainValidationStatus` missing `FAILED_TO_VERIFY` (P0)

The Okta API returns `"validationStatus": "FAILED_TO_VERIFY"` in domain responses (e.g. `GET /api/v1/domains/{domainId}`), but the SDK enum only declared `COMPLETED`, `IN_PROGRESS`, `NOT_STARTED`, `VERIFIED`. Jackson deserialization fell through to `UNKNOWN_DEFAULT_OPEN_API` via `READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE`, silently losing the original value. Added the missing constant to `src/swagger/api.yaml`.

### 2. `DevicePostureChecks.include` typed as `List<String>` (P0)

`DevicePostureChecks.include` had no `items` type in the spec - just an `example` - so the generator defaulted to `List<String>`, while the real API shape is a list of `{variableName, value}` pairs. Added a new `DevicePostureCheckMapping` schema and typed `include` as `List<DevicePostureCheckMapping>`.

### 3. CI fix: `ApplicationGrantsIT` non-existent-grant tests

Observed on `master` post-merge (unrelated to #1699's actual changes - a pre-existing latent test bug that CI happened to hit): `testGetNonExistentGrantError`/`testRevokeNonExistentGrantError` already acknowledge a `500`/`501` response as "OAuth grants not available" via an *outer* catch block, but that only covers exceptions thrown outside the inner try/catch. When `getScopeConsentGrant`/`revokeScopeConsentGrant` themselves returned `500`, the exception was caught into the local `exception` variable and then failed the "should return 404" assertion instead of being recognized as the same acknowledged case. Applied the same 500/501 check to that inner exception in both tests.

## Category
- [x] Bugfix
- [ ] Enhancement
- [ ] New Feature
- [ ] Library Upgrade
- [ ] Configuration Change
- [ ] Versioning Change
- [x] Unit or Integration Test(s)
- [ ] Documentation

## Signoff
- [ ] I have submitted a CLA for this PR (N/A - internal Okta contributor)
- [ ] I have updated documentation to explain what my PR does

### DIFF
--- a/api/src/test/java/com/okta/sdk/resource/model/DevicePostureChecksTest.java
+++ b/api/src/test/java/com/okta/sdk/resource/model/DevicePostureChecksTest.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2026-Present Okta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.okta.sdk.resource.model;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+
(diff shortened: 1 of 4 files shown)


aniket-okta (approved):
LGTM 🚀

PR #1707, opened 2026-08-18
Fix unbounded server-controlled retry sleep on HTTP 429 (OKTA-1235975)

prachi-okta:
## Summary

`OKTA-1235975`: the SDK's 429 retry delay was computed entirely from the server-supplied `x-rate-limit-reset` header with no upper bound, so a malicious or MitM server could force the client thread to sleep indefinitely. Since this sleep can happen inside `OAuth2ClientCredentials.applyToParams` (`synchronized`), one blocked call could stall every other caller waiting on token refresh — a client-side DoS triggerable by untrusted response data.

- `RetryUtil.get429DelayMillis` now clamps the computed delay to a `maxDelayMillis` ceiling.
- `OktaHttpRequestRetryStrategy` derives that ceiling from `retryMaxElapsed` if explicitly configured, otherwise falls back to the same 20s cap already used for 503/504/IOException retries — so 429 is no longer the exception.
- `DefaultClientBuilder` now actually wires `clientConfig.getRetryMaxElapsed()` into the strategy (previously set on `ClientConfiguration` but never read anywhere — dead config).
- README updated to document the new default cap and the existing `requestTimeout`/`setRetryMaxElapsed` escape hatch for anyone who needs a longer wait.

Separately (own commit, unrelated root cause): bumps `httpclient5` 5.6.2 → 5.6.4 to resolve **CVE-2026-64607** (connection-leak DoS in the classic I/O model on malformed `Content-Encoding` headers), found while chasing an unrelated stalled `release:prepare` run. This supersedes the existing `dependabot/maven/.../httpclient5-5.6.3` branch.

---
Have added more details of the fix in the comments: https://oktainc.atlassian.net/browse/OKTA-1235975

### DIFF
--- a/README.md
+++ b/README.md
@@ -723,6 +723,8 @@ okta.client.requestTimeout = 0 //Sets the maximum number of seconds to wait when
 okta.client.rateLimit.maxRetries = 4 //Sets the maximum number of attempts to retrying before giving up.
 ```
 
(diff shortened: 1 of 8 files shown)


aniket-okta (approved):
LGTM 🪨

